PR #1, opened 2019-04-02
get rid of prefixrouter.baseurl / readme cleanup

danielpoe:


framework/prefixrouter/module.go
```diff
@@ -80,31 +80,16 @@ func (m *Module) serve(
...
-
-			m.logger.WithField("category", "prefixrouter").Info("Routing ", area.Name, " at ", baseURL)
 			area.Injector.Bind((*flamingo.Logger)(nil)).ToInstance(m.logger.WithField("area", area.Name))
 			areaRouter := area.Injector.GetInstance(web.Router{}).(*web.Router)
-			prefix := areaRouter.Base().Host + areaRouter.Base().Path
-			// allow to override the baseurl
-			if baseurlVal, ok := area.Configuration.Get("prefixrouter.baseurl"); ok {
-				prefix = baseurlVal.(string)
+			prefix, loggingPrefix := getPrefixFromConfig(areaRouter)
+			if prefix == "/" {
```
thebod: `"/"` is fine, `""` (empty string) would be a problem

framework/prefixrouter/module.go
```diff
@@ -80,31 +80,16 @@ func (m *Module) serve(
...
-			}
-
-			if baseURL == "" {
-				m.logger.WithField("category", "prefixrouter").Warn("No baseurl configured for config area %v", area.Name)
-				continue
-			}
-
-			m.logger.WithField("category", "prefixrouter").Info("Routing ", area.Name, " at ", baseURL)
 			area.Injector.Bind((*flamingo.Logger)(nil)).ToInstance(m.logger.WithField("area", area.Name))
 			areaRouter := area.Injector.GetInstance(web.Router{}).(*web.Router)
```
thebod: We need to adjust this: as this is initialized after config.GetFlatContexts() we will fallthrough here, and child config areas will get their parents config. We need to fix that.

PR #9, opened 2019-04-11
WIP - generic auth port

danielpoe:


core/auth/module.go
```diff
@@ -0,0 +1,33 @@
+package auth
+
+import (
+	"flamingo.me/dingo"
+	interfaces2 "flamingo.me/flamingo/v3/core/auth/interfaces"
```
mathiasschopmans: AFAIS there's no need to use the alias.

```suggestion
	"flamingo.me/flamingo/v3/core/auth/interfaces"
```

core/auth/domain/auth.go
```diff
@@ -0,0 +1,78 @@
+package domain
+
+import (
+	"encoding/gob"
+	securityDomain "flamingo.me/flamingo/v3/core/security/domain"
+)
+
+type (
+	// Idendity information
+	Idendity interface {
```
thebod: Identity

core/auth/domain/auth.go
```diff
@@ -0,0 +1,78 @@
...
+	// User is a basic authenticated user
+	User interface {
+		Subject() string
+		Email() (string,bool)
+		Name() (string,bool)
+		CustomField(string) (string,bool)
+		Roles() []securityDomain.Role
+	}
+
+	//SimpleUser - implements User and can be used
```
thebod: As we will use godoc please format the comments like
```
// SimpleUser is a default implementation of the User interface...
```
and not
```
//SimpleUser - implements User ...
```
https://golang.org/doc/effective_go.html#commentary

core/auth/domain/auth.go
```diff
@@ -0,0 +1,78 @@
...
+	SimpleUser struct {
+		SubjectVal string
```
thebod: Run gofmt please

core/auth/domain/auth.go
```diff
@@ -0,0 +1,78 @@
...
+func (u *SimpleUser) Name() (string,bool) {
+	if u.NameVal == nil {
+		return "",false
+	}
+	return *u.NameVal, true
+}
+
+
+// CustomField - get a customfield by key
+func (u *SimpleUser) CustomField(key string) (string,bool) {
```
thebod: Simplify method by doing
```go
val, ok := u.CustomFieldVal[key]
return val, ok
```

core/auth/interfaces/controller.go
```diff
@@ -0,0 +1,48 @@
...
+// Inject for Authcontroller
+func (c *Authcontroller) Inject(responder *web.Responder, authservice Authservice,router web.ReverseRouter,) {
+	c.responder = responder
+	c.authservice = authservice
+	c.router = router
+}
+
+//AuthAction - default auth action - starting the authorsation with the registered Authservice
+func (c *Authcontroller) AuthAction(ctx context.Context, r *web.Request) web.Result {
+	redirecturl, ok := r.Params["redirecturl"]
```
thebod: You can safely omit the `ok` variable here, the default value for string is `""`

core/basicauth/interfaces/authservice.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+)
+
+var (
+	_ domain.Idendity = &Idendity{}
+)
+
+// Inject for Authservice
+func (o *Authservice) Inject(responder *web.Responder, config *struct {
+	Users config.Slice `inject:"config:basicauth.users,optional"`
```
thebod: Maybe users should be provided via something like a UserProvider? So the credentials do not have to be in the config?

core/basicauth/interfaces/authservice.go
```diff
@@ -0,0 +1,74 @@
...
+func (a *Authservice) Authenticate(ctx context.Context, returnURL *url.URL) (web.Result, error) {
+	redirectResponse := a.responder.URLRedirect(returnURL)
+	redirectResponse.Header.Add("WWW-Authenticate", `Basic realm="`+a.realm+`", charset="UTF-8"`)
+	redirectResponse.Status = http.StatusUnauthorized
+	return redirectResponse,nil
+}
+
+func (a *Authservice) IsAuthenticated(ctx context.Context, r *web.Request) bool {
+	_, _, ok := r.Request().BasicAuth()
+	//TODO check user pw
```
thebod: This needs to be done

core/auth/domain/auth.go
```diff
@@ -0,0 +1,78 @@
...
+		Roles() []securityDomain.Role
+	}
+
+	//SimpleUser - implements User and can be used
+	SimpleUser struct {
+		SubjectVal string
+		EmailVal *string
+		NameVal *string
+		CustomFieldVal map[string]string
+		DefaultRole string
```
thebod: What is the DefaultRole?

core/auth/domain/auth.go
```diff
@@ -9,32 +12,67 @@ type (
...
-		Street       string
-		ZipCode      string
-		City         string
-		DateOfBirth  string
-		Country      string
-		CustomFields map[string]string
-		Groups       []string
+	User interface {
+		Subject() string
+		Email() (string,bool)
```
Ompluscator: I would go with single return argument, with default value as empty string. In general it would be easier for further usage and easy to check for empty string if it's needed.

core/auth/interfaces/authservice.go
```diff
@@ -0,0 +1,19 @@
...
+	"context"
+	"net/url"
+	"flamingo.me/flamingo/v3/core/auth/domain"
+	"flamingo.me/flamingo/v3/framework/web"
+)
+
+type (
+	//Authservice - generic Authservice interface that should be used
+	Authservice interface {
+		Authenticate(ctx context.Context, returnURL *url.URL) (web.Result, error)
```
Ompluscator: As remainder to add web.Request in a list of arguments. We can omit to pass returnUrl - making of returnUrl is provided in security middleware, and it can be from header referrer or defined path in config (but, it doesn't hurt if we keep it).

core/auth/interfaces/controller.go
```diff
@@ -0,0 +1,48 @@
...
+
+import (
+	"context"
+	"errors"
+	"net/url"
+	"flamingo.me/flamingo/v3/framework/web"
+)
+
+type (
+	Authcontroller struct {
```
Ompluscator: Camel case: "AuthController"

core/auth/interfaces/controller.go
```diff
@@ -0,0 +1,48 @@
...
+		responder *web.Responder
+		authservice Authservice
```
Ompluscator: Camel cases and map of services: authServices map[string]AuthService

core/auth/interfaces/controller.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+//AuthAction - default auth action - starting the authorsation with the registered Authservice
+func (c *Authcontroller) AuthAction(ctx context.Context, r *web.Request) web.Result {
+	redirecturl, ok := r.Params["redirecturl"]
+	if !ok || redirecturl == "" {
+		redirecturl = r.Request().Referer()
+	}
+
+	if refURL, err := url.Parse(redirecturl); err != nil || refURL.Host != r.Request().Host {
```
Ompluscator: I'm not sure should we have this logic here: there is same functionality in security middleware, so we can either move that here completely or we just read "redirecturl" Param and pass url or nil as argument.

core/auth/interfaces/controller.go
```diff
@@ -0,0 +1,48 @@
...
+
+	if err != nil {
+		return c.responder.ServerError(errors.New("wrong redirect url given"))
+	}
+	result, err := c.authservice.Authenticate(ctx,url)
+	if err != nil {
+		return c.responder.ServerError(errors.New("wrong redirect url given"))
+	}
+	return result
+}
```
Ompluscator: Maybe we can add also Logout action?

core/basicauth/interfaces/authservice.go
```diff
@@ -0,0 +1,74 @@
+package interfaces
```
Ompluscator: Reminder: to have basicauth as module in separate repository.

bastianccm:
Superseded by #50

PR #14, opened 2019-04-16
Update date time formatter

chetan-thapliyal:
Add new method to override default time zone location.

### DIFF
--- a/core/locale/domain/dateTime.go
+++ b/core/locale/domain/dateTime.go
@@ -27,10 +27,12 @@ func (dtf *DateTimeFormatter) SetDateTime(time time.Time, localtime time.Time) {
 func (dtf *DateTimeFormatter) SetLocation(loc string) *DateTimeFormatter {
 	location, err := time.LoadLocation(loc)
 
-	if err != nil && dtf.logger != nil {
-		dtf.logger.WithField(flamingo.LogKeyMethod, "SetLocation").Error(
-			fmt.Sprintf("%s: failed to load time zone location: %v", loc, err),
-		)
+	if err != nil {
+		if dtf.logger != nil {
+			dtf.logger.WithField(flamingo.LogKeyMethod, "SetLocation").Error(
+				fmt.Sprintf("%s: failed to load time zone location: %v", loc, err),
+			)
+		}
 	} else {
 		dtf.localDateTime = dtf.dateTime.In(location)
 	}
--- a/core/locale/domain/dateTime.go
+++ b/core/locale/domain/dateTime.go
@@ -18,7 +18,8 @@ type DateTimeFormatter struct {
 	dateTime       time.Time
 }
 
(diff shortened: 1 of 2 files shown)


core/locale/domain/dateTime.go
```diff
@@ -17,6 +23,27 @@ func (dtf *DateTimeFormatter) SetDateTime(time time.Time, localtime time.Time) {
 	dtf.localDateTime = localtime
 }
 
+// SetLocation sets location for local time
+func (dtf *DateTimeFormatter) SetLocation(loc string) *DateTimeFormatter {
+	location, err := time.LoadLocation(loc)
+
+	if err != nil && dtf.logger != nil {
```
carstendietrich: Condition needs to be split, if there is an error and no logger the invalid location is still used which leads to a panic

chetan-thapliyal: You are right. It has been fixed.

core/locale/domain/dateTime.go
```diff
@@ -17,6 +23,29 @@ func (dtf *DateTimeFormatter) SetDateTime(time time.Time, localtime time.Time) {
...
 }
 
+// SetLocation sets location for local time
+func (dtf *DateTimeFormatter) SetLocation(loc string) *DateTimeFormatter {
+	location, err := time.LoadLocation(loc)
+
+	if err != nil {
+		if dtf.logger != nil {
+			dtf.logger.WithField(flamingo.LogKeyMethod, "SetLocation").Error(
+				fmt.Sprintf("%s: failed to load time zone location: %v", loc, err),
```
masmrlar: Why exactly logging here and skip to report the error to the callier? 
I would appreciate to report back the error occouring on setting an invalid time location - the caller could then decide to go on or handle/abort.

By just logging no code using this would recognize that SetLocation is not successful

core/locale/domain/dateTime.go
```diff
@@ -17,6 +23,29 @@ func (dtf *DateTimeFormatter) SetDateTime(time time.Time, localtime time.Time) {
...
+		}
+	} else {
+		dtf.localDateTime = dtf.dateTime.In(location)
+	}
+
+	return dtf
+}
+
+// SetLogger sets the logger instance
+func (dtf *DateTimeFormatter) SetLogger(logger flamingo.Logger) {
```
masmrlar: With my comments above we won't need a logger in here.

core/locale/domain/dateTime.go
chetan-thapliyal: I updated the code to return an error. I left the logging with original error as error description also acts as an error code/key in FE context.

core/locale/domain/dateTime.go
Ompluscator: I agree. I think logger should not be a part of DateTimeFormatter. Method "SetLocation" already returns an error, if there is one, so it can be logged from the place when the method is called. :)

PR #15, opened 2019-04-16
prefixrouter: fixed issue related to ending slashes to support paths …

Ompluscator:
…with or without ending slashes


thebod:
You can not use `path.Join` on Request URI's! this will remove all duplicated `/` characters in GET parameters!

PR #21, opened 2019-04-17
added unit test support for go 1.11 and 1.12 via travis-ci

Ompluscator:
travis-ci kickstart

fixing go.mod

fixing go.mod

finalizing travis.ci


thebod:
We should add strict golint here, and run go test with `go test -vet=all ./...`

PR #25, opened 2019-04-25
Split Readme and add method to publish multiple events

danielpoe:


thebod:
pls squash before merge

PR #26, opened 2019-04-25
add fine grained cache control header support

danielpoe:
Currently there is only a SetNoCache Helper for responses

Issues:
- No default cache control strategy can be set
- The current "SetNoCache" - is not 100% correct 
- There is always deeper caching know how involved for setting correct cache related http headers - we should provide a consitent and easy way to set and only "build" reasonable header combinations


framework/web/result.go
```diff
@@ -75,16 +78,44 @@ type (
 		RenderResponse
 		Error error
 	}
+
+	//CacheControlHeader - holds the possible directives for Cache Control Headers
```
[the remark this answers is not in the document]
danielpoe: In the method SetHeaders

no-cache and ETag is not mutually exclusive by the way

no-store and ETag is

framework/web/result.go
```diff
@@ -384,3 +429,114 @@ func (r *Responder) getLogger() flamingo.Logger {
...
+	if c.Visibility == "private" {
+		cacheControlValues = append(cacheControlValues, "private")
+	}
+
+	if len(cacheControlValues) > 0 {
+		header.Set("Cache-Control", strings.Join(cacheControlValues, ", "))
+	}
+}
+
+//NewCacheStrategy returns new cache Strategy
```
thebod: This factory is not necessary, and actually not always used, e.g. by the Filter.

framework/web/filter/cachestrategy.go
```diff
@@ -0,0 +1,34 @@
...
+
+import (
+	"context"
+	"flamingo.me/flamingo/v3/framework/web"
+	"net/http"
+)
+
+type (
+	//CacheStrategy - a filter that sets CacheDirective if not present
+	CacheStrategy struct {
```
thebod: Make this type private

framework/web/filter/cachestrategy.go
```diff
@@ -0,0 +1,34 @@
...
+	response := chain.Next(ctx, r, w)
+	if defaultResponse, ok := response.(*web.Response); ok {
+		if defaultResponse.CacheDirectives == nil && r.Request().Method == http.MethodGet {
+			cacheStrategy := web.CacheStrategy{}
+			cacheStrategy.SetIsReusable(f.DefaultIsReuseable)
+			cacheStrategy.SetRevalidateEachTime(f.DefaultRevalidateEachTime)
+			cacheStrategy.SetAllowIntermediateCaches(f.DefaultAllowIntermediateCaches)
+			cacheStrategy.SetMaxCacheLifetime(int(f.DefaultMaxCacheLifetime))
+			defaultResponse.CacheDirectives = cacheStrategy.Build()
+			return defaultResponse
```
thebod: Return the `response` variable. The `defaultResponse` does not contain everything, you basically unpack the `web.Response` from the actual `web.Result`.

framework/web/result.go
```diff
@@ -75,16 +77,58 @@ type (
...
+	CacheStrategy struct {
+		isReusable bool
+		revalidateEachTime bool
+		allowIntermediateCaches bool
+		maxCacheLifetime int
+		etag string
+	}
+
+	//CacheDirectives - holds the possible directives for Cache Control Headers and other Http Caching
+	CacheDirectives struct {
```
thebod: `CacheDirective` singular

framework/web/filter/cachestrategy.go
```diff
@@ -17,18 +18,26 @@ type (
...
-			cacheStrategy.SetAllowIntermediateCaches(f.DefaultAllowIntermediateCaches)
-			cacheStrategy.SetMaxCacheLifetime(int(f.DefaultMaxCacheLifetime))
-			defaultResponse.CacheDirectives = cacheStrategy.Build()
-			return defaultResponse
-		}
+	if r.Request().Method != http.MethodGet {
+		return response
+	}
+	switch typedResponse := response.(type) {
+	case *web.RenderResponse:
```
thebod: Is this supposed to be `*web.RenderResponse`? You might want to use `*web.Response`?

PR #28, opened 2019-05-05
Commands not stopping

danielpoe:
Currently all commands are not stopping unless you hit STRG-C

This is because the rootcommand inherits the PersistentPostRun which always blocks until termination signal is send.

Proposal A) 
We can send own signal in all commands that should stop immediatelly (see commit)

Proposal B)
The root command should not block
Instead it should just trigger the ShutdownEvent in its PersistentPostRun
Then it is the job of the "serve" command to handle the interruption

Thoughts?

### DIFF
--- a/framework/cmd/module.go
+++ b/framework/cmd/module.go
@@ -64,7 +64,7 @@ func (m *Module) Configure(injector *dingo.Injector) {
 
 			return rootCmd
 		},
-	)
+	).In(dingo.Singleton)
 }
 
 // DefaultConfig specifies the command name


tessig:
The PersistentPostRun is just to trigger the graceful shutdown. The `flamingo.ShutdownEvent` is dispatched and then we wait for all modules to handle their shutdown processes with a timeout of 30 seconds.

tessig:
I think it's just a race condition between `shutdown()` in framework/cmd/module.go:76 and `PersistentPostRun` of the root command.

thebod:
We should not use signals to signal ourself. Actually we should not signal ourselves at all, as this is a feature which is totally unnecessary.

The default command behaviour should work different:
- main routine doing the commands main work, such as running the command or server or whatever
- one goroutine checking for signals, such as sigint & co. this also switches on a second channel or whatever to be notified if the main process is done.

This way the signal processing is just a second mechanism to terminate, but should never be blocking

tessig:
ok, the root cause is, that the dingo provider for the root command is called multiple times, which leads to have multiple `signals` channels and only one receives the signal while all others are blocking.
moving the channel variable to global package space would only hide the problem.
So the most acceptable fix for now would be to define the root command as singleton.

thebod:
@danielpoe @tessig can you retest? then we can merge it.
i rebased on master

tessig:
@thebod rebased another time and tested ok.

danielpoe:
@thebod can this be merged?

PR #33, opened 2019-05-10
handle accesstoken empty edge cases

danielpoe:
Observations:
* Accesstoken seems to be empty in some cases (e.g. error in oauth2 provider)

Other Issues:
* The openid token is saved only once in callback controller in the session
* services that need to get the accesstoken do this normaly by:

```
1 auth = authmanager.Auth()
2 token = auth.TokenSource.Token()
3 token.AccessToken
```
1 gets tokensource from saved oauth2 token in session
   - how does that work since a reference is supposed to be saved in session:
         - after writing the session?
         - in cluster setup?
         (imho pointer should never be saved in session)

2  this call will renew the token if it is expired - but never saved back to session

Solution:
* Whenever a consumer calls  authmanager.Auth() - he should get is based on an eventually renewed "new" Token
* write no pointer in session

### DIFF
--- a/core/oauth/Readme.md
+++ b/core/oauth/Readme.md
@@ -103,3 +103,8 @@ html
   ...
   a(href=url("auth.callback")) Login
 ```
+
+
+## Debugging
+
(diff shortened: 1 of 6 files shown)


thebod:
@danielpoe can you please rebase and remove the merge commits?

PR #46, opened 2019-07-10
Version endpoint

danielpoe:
Some projects want build time informations returned by a path.

Typically this is done by writing a version.json file during build time.

This Feature provides an endpoint to return the content of that file, if given.

### DIFF
--- a/core/healthcheck/Readme.md
+++ b/core/healthcheck/Readme.md
@@ -2,7 +2,8 @@
 
 The healthcheck module provides useful routes for:
 1. check if the application is up (/status/ping endpoint)
-1. Check the application status (/status/healthcheck endpoint)
+2. Check the application status (/status/healthcheck endpoint)
+3. Provide version informations (/version) - if a file "version.json" exists
 
 ## Usage
 
--- a/core/healthcheck/interfaces/controllers/version.go
+++ b/core/healthcheck/interfaces/controllers/version.go
@@ -1,0 +1,32 @@
+package controllers
+
+import (
+	"net/http"
+
+)
+
+type (
+
+	// Version controller
+	Version struct {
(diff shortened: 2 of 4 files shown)


thebod:
We should check "go build -X..." To set version info during compilation time.

panzerfahrer:
Is this really worth putting into the framework? Feels slightly like the start of _eierlegendewollmilchsau_, doesn't it?

danielpoe:
> We should check "go build -X..." To set version info during compilation time.

I like this idea.

Not sure about "eierlegendewollmilchsau" - one the one hand yes - on the other a quite common and simple use case and the code does not add complexity

panzerfahrer:
This looks just like quite an opinionated solution and in the context of "some projects want" I'm not sure if this really adds value to the framework.

Hooking into the `SystemServer` is basically a 5-liner. And if that's quick enough to do, framework users can just add what ever they want (version numbers, git sha, timestamps, ...).

So I'd suggest to rather add method that allows to extend the behavior of the `SystemServer`.

danielpoe:
> This looks just like quite an opinionated solution and in the context of "some projects want" I'm not sure if this really adds value to the framework.
> 
> Hooking into the `SystemServer` is basically a 5-liner. And if that's quick enough to do, framework users can just add what ever they want (version numbers, git sha, timestamps, ...).
> 
> So I'd suggest to rather add method that allows to extend the behavior of the `SystemServer`.

by returning the content of a variable "BUILDINFO" or whatever it returns what everyone wants to be returned.
I still dont see that this feature is a super project specific thing but rather something generic

bastianccm:
Alternative proposal in #156

bastianccm:
For now superseded by #156

PR #48, opened 2019-07-17
oauth: support multiple claim keys in mapping config

danielpoe:
There are cases where properties need to be mapped from different claim keys.
This change propose to configure multiple fields with ";" seperated

### DIFF
--- a/core/oauth/domain/mapping.go
+++ b/core/oauth/domain/mapping.go
@@ -125,9 +125,9 @@ func (ums *UserMappingService) mapCustomFields(mapping []string, claims map[stri
 }
 
 func (ums *UserMappingService) mapField(mappedFieldName string, claims map[string]interface{}) string {
-	claimKeys := strings.Split(mappedFieldName,";")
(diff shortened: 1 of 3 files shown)


core/oauth/domain/mapping.go
```diff
@@ -125,11 +125,13 @@ func (ums *UserMappingService) mapCustomFields(mapping []string, claims map[stri
 }
 
 func (ums *UserMappingService) mapField(mappedFieldName string, claims map[string]interface{}) string {
-	value, ok := claims[mappedFieldName].(string)
-	if !ok {
-		return ""
+	claimKeys := strings.Split(mappedFieldName,";")
+	for _,key := range claimKeys {
+		if value, ok := claims[key].(string); ok {
```
thebod: Please use `strings.TrimSpace(key)` to ensure that whitespaces around the key are ignored

carstendietrich: @thebod Updated the code accordingly, please review and merge.

PR #50, opened 2019-07-23
create new auth module

bastianccm:
As discussed some time ago, the interfaces and a HTTP Basic Auth example for a possible integration


core/auth/identifier.go
```diff
@@ -0,0 +1,51 @@
...
+	injector.BindMulti(new(RequestIdentifier)).To(new(HTTPBasicAuthIdentifier))
+}
+
+// HTTPBasicAuthIdentifier identifies users based on HTTP Basic Authentication header
+type HTTPBasicAuthIdentifier struct {
+	users map[string]string
+}
+
+// Inject user configuration
+func (i *HTTPBasicAuthIdentifier) Inject(config *struct {
```
danielpoe: identifier contains a sample implementation for http basic auth:
- should we rename file to "basicauth" then? 
-or place this in a subpackage/subfolder "basicauth" to also differenciate it from the other types in that module (that really belongs to the generic auth core)

danielpoe:
What did we discuss about adjusting security module?

At least we should also add a role provider that is using the configured idendity - so that security module works out of the box

danielpoe:
The basic auth example cannot be used out of the box - since the "authenticate" part is missing where we send the header that makes the browser open the user/pw dialog.

Maybe we should then move the basicauth stuff to own module and provide also controller for this that gets redirect url to redirect to after sucessful authentication.. This way it acts as a "full" but easy example

danielpoe:
We should also add a Readme, explaining how the module can be used (check Idendity) and how to implement own Idendificators


The docu should also explain the more advances usage of the security module in combination with auth - to protect actions by either:
 - wrap in bindings or
 - explicit call role provider

core/auth/identity.go
```diff
@@ -0,0 +1,26 @@
+package auth
+
+type (
+	// Identity donates an authentication object which at least identifies the authenticated subject
+	Identity interface {
+		Subject() string
```
danielpoe: We are not providing a generic "User" struct anymore as a simple unified model for all other modules requiring a simple "user" ?
That would mean every other auth implementation need to add this?

(Maybe as optional extra interface like the "HasRoles" one)

core/auth/webidentity.go
```diff
@@ -0,0 +1,43 @@
...
+import (
+	"context"
+
+	"flamingo.me/flamingo/v3/framework/web"
+)
+
+type (
+	// RequestIdentifier resolves an identity from a web request
+	RequestIdentifier interface {
+		Identify(ctx context.Context, request *web.Request) Identity
```
danielpoe: unauthenticated request would return nil?

That would then in future be the way how to check for logged in or not?

core/auth/identifier.go
```diff
@@ -0,0 +1,51 @@
...
+	Users config.Map `inject:"config:core.auth.httpbasicusers"`
+}) *HTTPBasicAuthIdentifier {
+	config.Users.MapInto(&i.users)
```
golangcibot: Error return value of `config.Users.MapInto` is not checked (from `errcheck`)

core/auth/identity.go
bastianccm: What would a generic "User" be like?

core/auth/webidentity.go
bastianccm: Right now: yes, though we could return an error or bool here too?
Also for guests: we can return a GuestIdentity or nil.

core/auth/identity.go
danielpoe: something like User: email, firstname,lastname as a bare minimum

core/auth/controller.go
```diff
@@ -0,0 +1,36 @@
+package auth
+
+import (
+	"context"
+
+	"flamingo.me/flamingo/v3/framework/web"
+)
+
+type WebDecorator struct {
```
Ompluscator: Should this be something like middleware? If yes, I would maybe rename it so, just for purpose of consistency (we already use that naming, and it's a kind of standard in routing go frameworks).

core/auth/controller.go
```diff
@@ -0,0 +1,36 @@
...
+	wis *WebIdentityService
+}
+
+func (wd *WebDecorator) RequireIdentified(action web.Action) web.Action {
```
Ompluscator: Probably we can remove some methods from security module inside app service and middleware, which are doing similar job.

core/auth/webidentity.go
Ompluscator: As I remember - idea was to have role provider in auth module, which would add PermissionAuthorised (or similar), and base on that we should now if user is logged in or not -> so, from that perspective, checking if Identity is nil or by additional error/bool value, would be same :)

core/auth/webidentity.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+
+	// WebAuthenticater returns the first usable web.Result to return to authenticate the user
+	WebAuthenticater interface {
+		Authenticate(ctx context.Context, request *web.Request) web.Result
+	}
+
+	// WebIdentityService calls one or more identifier to get all possible identities of a user
+	WebIdentityService struct {
```
Ompluscator: Maybe we may add possibility to set default identityProvider in config - in general, anywhere where there are multiple possibilities for authentication, default login page is shown (like keycloak or any other), beside having additional posibilities to login via facebook/google. So, maybe adding IdentifyForDefault/AuthenticateForDefault would be useful.

core/auth/identity.go
bastianccm: We can not assume that this information will exist. HTTP auth for example only provides a username/identifier, nothing more. Also we had projects in the past where no user information but a technical ID was available. So, unless there is a standard which defines names, naming scheme, etc., we can not use it.

Also firstname and lastname will not apply for machine authentication.

core/auth/webidentity.go
bastianccm: Yes, security has the same, i might remove some here

core/auth/controller.go
bastianccm: Yes, middleware, didn't remember the wording -.-*

core/auth/module.go
```diff
@@ -0,0 +1,90 @@
...
+}
+
+func buildAuthentifier(
+	provider map[string]IdentifierFactory,
+	cfg *struct {
+		Config config.Slice `inject:"config:core.auth.web.broker"`
+	},
+) []RequestIdentifier {
+	var broker []config.Map
+	cfg.Config.MapInto(&broker)
```
golangcibot: Error return value of `cfg.Config.MapInto` is not checked (from `errcheck`)

core/auth/module.go
```diff
@@ -0,0 +1,90 @@
...
+}) {
+	r.debugController = debugController
+	r.controller = controller
+	r.debug = cfg.Debug
+}
+
+// Routes configuration
+func (r *routes) Routes(router *web.RouterRegistry) {
+	if r.debug {
+		router.Route("/core/auth/debug", "core.auth.debug")
```
golangcibot: Error return value of `router.Route` is not checked (from `errcheck`)

core/auth/module.go
```diff
@@ -0,0 +1,90 @@
...
+		router.HandleAny("core.auth.debug", r.debugController.Action)
+	}
+	router.Route("/core/auth/callback/:broker", "core.auth.callback(broker)")
```
golangcibot: Error return value of `router.Route` is not checked (from `errcheck`)

core/auth/module.go
```diff
@@ -0,0 +1,90 @@
...
+	router.HandleAny("core.auth.callback", r.controller.Callback)
+	router.Route("/core/auth/login/:broker", "core.auth.login(broker)")
```
golangcibot: Error return value of `router.Route` is not checked (from `errcheck`)

core/auth/http/http.go
```diff
@@ -0,0 +1,69 @@
...
+// basicAuthIdentifier identifies users based on HTTP Basic Authentication header
+type basicAuthIdentifier struct {
+	users  map[string]string
+	realm  string
+	broker string
+}
+
+func identifierFactory(cfg config.Map) auth.RequestIdentifier {
+	i := new(basicAuthIdentifier)
+	config.Map(cfg["users"].(map[string]interface{})).MapInto(&i.users)
```
golangcibot: Error return value of `(flamingo.me/flamingo/v3/framework/config.Map).MapInto` is not checked (from `errcheck`)

core/auth/oauth/oidc.go
```diff
@@ -0,0 +1,238 @@
...
+
+	if err != nil {
+		panic(err)
+	}
+
+	if idtoken, ok := token.Extra("id_token").(string); ok {
+		i.rawIDToken = idtoken
+	}
+
+	idToken, err := i.verifier.Verify(ctx, i.rawIDToken)
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

core/auth/webidentity.go
bastianccm: First one is default

core/auth/identity.go
```diff
@@ -0,0 +1,52 @@
...
+		Broker() string
+	}
+
+	// hasRoles adds the ability to provide roles to an identity
+	hasRoles interface {
```
golangcibot: type `hasRoles` is unused (from `unused`)

core/auth/webidentity.go
```diff
@@ -0,0 +1,208 @@
...
+	for _, provider := range s.identityProviders {
+		if identity, _ := provider.Identify(ctx, request); identity != nil {
+			identities = append(identities, identity)
+		}
+	}
+
+	return identities
+}
+
+func (s *WebIdentityService) storeRedirectURL(request *web.Request) {
```
bastianccm: This does not work properly for logins on the current page, without redirecting to /core/auth/login

core/oauth/example/main.go
```diff
@@ -0,0 +1,62 @@
...
+type routes struct {
+	c *controller
+}
+
+func (r *routes) Inject(c *controller) {
+	r.c = c
+}
+
+func (r *routes) Routes(registry *web.RouterRegistry) {
+	registry.Route("/", "example")
```
golangcibot: Error return value of `registry.Route` is not checked (from `errcheck`)

PR #54, opened 2019-07-29
prefixrouter: use longest prefix if we have ambiguous matches

tessig:
On ambiguous matches the selected route should be deterministic.
The old behaviour using a map range leads to a random routing.


thebod:
This change is not about deterministic behaviour but about always preferring the longest prefix.

PR #63, opened 2019-08-25
locale: add price service for formatting

danielpoe:
flamingo_commerce introduced a priceService to format prices and allows to configure formatting per currency.
This should move to locale package.

(later flamingo_commerce can use this service instead of own)

### DIFF
--- a/core/locale/module.go
+++ b/core/locale/module.go
@@ -34,15 +34,17 @@ func (m *Module) DefaultConfig() config.Map {
 		"locale": config.Map{
 			"locale": "en-US",
 			"accounting": config.Map{
-				"decimal":    ",",
-				"thousand":   ".",
-				"formatZero": "%s -,-",
-				"format":     "%s %v",
-				"formatLong": "%v %v",
+				"default": config.Map{
+					"decimal":    ".",
+					"thousand":   ",",
+					"formatZero": "%s 0.00",
+					"format":     "%s %v",
+					"formatLong": "%v %v",
+				},
 			},
 			"numbers": config.Map{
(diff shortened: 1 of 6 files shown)


core/locale/module.go
```diff
@@ -34,11 +34,13 @@ func (m *Module) DefaultConfig() config.Map {
...
 			"accounting": config.Map{
-				"decimal":    ",",
-				"thousand":   ".",
-				"formatZero": "%s -,-",
-				"format":     "%s %v",
-				"formatLong": "%v %v",
+				"default": config.Map{
+					"decimal":    ",",
+					"thousand":   ".",
+					"formatZero": "%s -,-",
```
thebod: Should this be `%s.-` instead of `%s -,-`?

core/locale/module.go
```diff
@@ -34,11 +34,13 @@ func (m *Module) DefaultConfig() config.Map {
 		"locale": config.Map{
 			"locale": "en-US",
 			"accounting": config.Map{
-				"decimal":    ",",
-				"thousand":   ".",
-				"formatZero": "%s -,-",
-				"format":     "%s %v",
-				"formatLong": "%v %v",
+				"default": config.Map{
+					"decimal":    ",",
```
thebod: The default locale is `en-US`, so this should be `.`, right? As well as `,` for the thousands operator default

danielpoe: yes makes sense - going to change

core/locale/module.go
danielpoe: adjusted

PR #69, opened 2019-08-29
add nancy vulnerability check to pipeline

tessig:


.travis.yml
```diff
@@ -9,11 +9,18 @@ env:
 
 install: true
 
+before_script:
+  - cd /tmp && mkdir tools && cd -
+  - sudo curl -s -L "https://github.com/sonatype-nexus-community/nancy/releases/download/v0.0.21/nancy-linux.amd64-v0.0.21" -o "/tmp/tools/nancy"
```
thebod: why sudo?

.travis.yml
```diff
@@ -9,11 +9,18 @@ env:
...
+  - export PATH=$PATH:/tmp/tools/
+
 script:
   - set -e
   - fmt=$(gofmt -l -s .)
   - test -z $fmt || (echo "please run gofmt" ; echo $fmt ; exit 1)
   - go get golang.org/x/lint/golint
   - golint -set_exit_status $(go list ./...)
   - go test -race -vet=all ./...
+  - nancy go.sum
```
thebod: if this is faster than go test move it up, earlier results are always better

.travis.yml
tessig: copied from the official example, but you're right it's not necessary

tessig:
now we have found a vulnerability

thebod:
Blocked by sonatype-nexus-community/nancy/issues/24

DarthHater:
Thanks for using Nancy! I replied to your issue over there, we can probably take some time to modify Nancy to take the output of `go list` now that it's in 1.13, and use that instead of `go.sum`. Pop on over and maybe we can work on it together!

PR #80, opened 2019-10-14
cache: test HTTPFrontend

tessig:
a test to cover the most common cases for HTTPFrontend


core/cache/httpFrontend_test.go
```diff
@@ -0,0 +1,281 @@
...
+					lifetime:  inPast,
+					gracetime: inFuture,
+				},
+				Data: cachedResponse{
+					orig: createResponse(200, "foo"),
+					body: []byte("foo"),
+				},
+			},
+			// we expect the cached value as result, but a the actual value from loader to be cached (async)
+			want:             createResponse(200, "foo"),
```
panzerfahrer: Isn't this too inconsistent - why wouldn't you want to use the latest response?

tessig: getting the latest response can take a lot of time. We want to answer quickly with the old response from the cache and assure that the next requests get the new result though

PR #83, opened 2019-10-24
core/gotemplate: Add template function tests and fix array out of bounds

carstendietrich:
This adds tests for the templatefucntions used in the gotemplate module

### DIFF
--- a/core/gotemplate/templatefunctions_test.go
+++ b/core/gotemplate/templatefunctions_test.go
@@ -6,6 +6,7 @@ import (
 	"html/template"
 	"net/url"
 	"reflect"
+	"sort"
 	"testing"
 )
 
@@ -16,7 +17,14 @@ func (m MockRouter) Relative(name string, params map[string]string) (*url.URL, e
 }
 
 func (m MockRouter) Data(ctx context.Context, handler string, params map[interface{}]interface{}) interface{} {
-	return fmt.Sprintf("%v %v", handler, params)
+	var stringParams []string
+	for key, value := range params {
+		stringParams = append(stringParams, fmt.Sprintf("%v:%v", key, value))
+	}
+
+	sort.Strings(stringParams)
+
+	return fmt.Sprintf("%v %v", handler, stringParams)
 }
 
 var _ urlRouter = MockRouter{}
@@ -37,7 +45,7 @@ func Test_dataFunc_Func(t *testing.T) {
 			args: args{
 				what: "test",
 			},
-			want: "test map[]",
+			want: "test []",
 		},
 		{
 			name: "valid params set",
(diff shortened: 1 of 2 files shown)


core/gotemplate/templatefunctions_test.go
```diff
@@ -0,0 +1,220 @@
...
+
+			templateFunction := gf.Func(context.Background()).(func(what string, params ...string) interface{})
+			if got := templateFunction(tt.args.what, tt.args.params...); !reflect.DeepEqual(got, tt.want) {
+				t.Errorf("getFunc() = %q, want %q", got, tt.want)
+			}
+		})
+	}
+}
+
+func Test_plainHTMLFunc_Func(t *testing.T) {
```
thebod: Do not use a complete table test for one single test case, this is too complex to maintain

core/gotemplate/templatefunctions_test.go
```diff
@@ -0,0 +1,220 @@
...
+			phf := &plainHTMLFunc{}
+			templateFunction := phf.Func(context.Background()).(func(in string) template.HTML)
+			if got := templateFunction(tt.in); !reflect.DeepEqual(got, tt.want) {
+				t.Errorf("plainHTMLFunc() = %q, want %q", got, tt.want)
+			}
+		})
+	}
+}
+
+func Test_plainJSFunc_Func(t *testing.T) {
```
thebod: Same, too complex testcase for a simple test.

core/gotemplate/templatefunctions_test.go
```diff
@@ -0,0 +1,220 @@
...
+	"context"
+	"fmt"
+	"html/template"
+	"net/url"
+	"reflect"
+	"sort"
+	"testing"
+)
+
+type MockRouter struct{}
```
thebod: Do not export MockRouter (make it lowercase)

core/gotemplate/templatefunctions_test.go
```diff
@@ -0,0 +1,220 @@
...
+
+			templateFunction := df.Func(context.Background()).(func(what string, params ...string) interface{})
+			if got := templateFunction(tt.args.what, tt.args.params...); !reflect.DeepEqual(got, tt.want) {
+				t.Errorf("dataFunc() = %q, want %q", got, tt.want)
+			}
+		})
+	}
+}
+
+func Test_getFunc_Func(t *testing.T) {
```
thebod: This test is the same as `Test_dataFunc_Func`, with the only difference in calling `get` instead of `data`, which is just an alias for `data`.
Remove this test and call both `get` and `data` in the `Test_dataFunc_Func` function, otherwise these tests won't be maintainable (duplicate code).

carstendietrich:
Hey @bastianccm,

thanks for the detailed feedback, I adjusted the tests accordingly and abolished the table tests, as they were in fact overkill.

PR #84, opened 2019-10-26
core/gotemplate: Add engine tests, unlock mutex in case of error

carstendietrich:
Add tests for the engine
Fix mutex staying locked in case of error which can lead to a deadlock
Fix template function `map` causing out of bounds error if called with not enough arguments

### DIFF
--- a/core/gotemplate/engine.go
+++ b/core/gotemplate/engine.go
@@ -66,9 +66,7 @@ func (e *engine) Render(ctx context.Context, name string, data interface{}) (io.
 
 	lock.Lock()
 	if e.debug || e.templates == nil {
-
 		err := e.loadTemplates(ctx)
-
 		if err != nil {
 			lock.Unlock()
 			return nil, err
--- a/core/gotemplate/engine.go
+++ b/core/gotemplate/engine.go
@@ -64,14 +64,17 @@ func (e *engine) Render(ctx context.Context, name string, data interface{}) (io.
 	ctx, span := trace.StartSpan(ctx, "gotemplate/Render")
 	defer span.End()
 
+	lock.Lock()
 	if e.debug || e.templates == nil {
-		lock.Lock()
+
 		err := e.loadTemplates(ctx)
-		lock.Unlock()
+
 		if err != nil {
+			lock.Unlock()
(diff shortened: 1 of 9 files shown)


core/gotemplate/engine.go
```diff
@@ -64,14 +64,14 @@ func (e *engine) Render(ctx context.Context, name string, data interface{}) (io.
 	ctx, span := trace.StartSpan(ctx, "gotemplate/Render")
 	defer span.End()
 
-	lock.Lock()
 	if e.debug || e.templates == nil {
+		lock.Lock()
```
thebod: This will lead to a race condition when e.templates is nil, eventually triggering e.loadTemplates multiple times. Move lock.Lock() up, and call lock.Unlock() both in error and non-error case.

core/gotemplate/engine_test.go
```diff
@@ -0,0 +1,183 @@
...
+	return make(map[string]flamingo.TemplateFunc)
+}
+
+var additionalTemplateFuncs = func() map[string]flamingo.TemplateFunc {
+	funcs := make(map[string]flamingo.TemplateFunc)
+	funcs["customTemplateFunc"] = CustomTemplateFunc{}
+	return funcs
+}
+
+type CustomTemplateFunc struct{}
```
thebod: Don't export

core/gotemplate/engine_test.go
```diff
@@ -0,0 +1,183 @@
...
+		debug              bool
+		tplFuncs           func() map[string]flamingo.TemplateFunc
+	}
+	type args struct {
+		name string
+		data interface{}
+	}
+	tests := []struct {
+		name    string
+		fields  fields
```
thebod: It is not clear what `fields` and `args` mean, maybe rename fields to engineConfig and args to renderArgs?

thebod:
Unittests in 1.11 are failing, despite restart

carstendietrich:
Hey @bastianccm,

fixed the mentioned stuff and switched to go-cmp instead of reflect.DeeplEqual. Tests in Go 1.11.x are now running as well. Please have a look.

ty

PR #85, opened 2019-11-04
accounting: update accounting library for bugfix

bykof:
The accounting library had an error regarding zeroFormat. It's fixed with the new version.


bastianccm:
@bykof can you amend and push with the correct go.sum? CI won't let this pass otherwise.

bykof:
@bastianccm done :>

PR #86, opened 2019-11-10
move httpFrontent_test.go to cache_test package

jochen42:
This is just an cosmetic pr. To be able to move HttpFrontent_test.go from the cache-package to the cache_test-packed, i needed to do some small refactorings:

- factory cache.NewEntryMeta: needed to build an Meta-Object inside tests
- exported cache.cachedResponse struct: needed to convert an interface to cache.CachedRepsonse conversion inside the tests
- factory cache.NewCachedResponse: needed to build an CachedResponse inside tests
- getter cache.CachedResponse.Body(): needed to compare the cache-results inside tests

As far as i can see, the changes are backward-compatible and hotfix-version compatible (correct me if i am wrong ;)).

Just cosmetic, but from my point of view unit-tests should be always in additional packages.

### DIFF
--- a/core/cache/cache.go
+++ b/core/cache/cache.go
@@ -36,3 +36,12 @@ type (
 		span trace.SpanContext
 	}
 )
+
+// NewEntryMeta builds a new Meta Object for an CacheEntry
+func NewEntryMeta(lifetime time.Time, gracetime time.Time, tags []string) Meta {
+	return Meta{
+		lifetime:  lifetime,
+		gracetime: gracetime,
+		Tags:      tags,
+	}
+}
--- a/core/cache/httpFrontend.go
+++ b/core/cache/httpFrontend.go
@@ -29,11 +29,25 @@ type (
 		io.Reader
 	}
 
-	cachedResponse struct {
+	// CachedResponse represents an cache http response entry
+	CachedResponse struct {
 		orig *http.Response
 		body []byte
 	}
 )
+
+// NewCachedResponse creates an new repsonse object for http frontend cache response
+func NewCachedResponse(orig *http.Response, body []byte) CachedResponse {
+	return CachedResponse{
+		orig: orig,
+		body: body,
+	}
+}
+
+// Body is an getter for cachedResponse.body
+func (cb CachedResponse) Body() []byte {
+	return cb.body
+}
 
 // Inject HTTPFrontend dependencies
 func (hf *HTTPFrontend) Inject(backend Backend, logger flamingo.Logger) *HTTPFrontend {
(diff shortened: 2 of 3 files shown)


bastianccm:
In general I'd vote against this, because that exposes several internal mechanisms such as creating arbitrary cache metadata, creating cachedResponses, etc.

Using test-packages is useful if you want to test explicit the exposed API. However this package has a lot of internal stuff, exposing that would cause additional API to maintain.

After all one is not supposed to create cache entries by hand, so, unless we want someone to do that, we shouldn't expose these functions.

What do you think?

jochen42:
> In general I'd vote against this, because that exposes several internal mechanisms such as creating arbitrary cache metadata, creating cachedResponses, etc.
> 
> Using test-packages is useful if you want to test explicit the exposed API. However this package has a lot of internal stuff, exposing that would cause additional API to maintain.
> 
> After all one is not supposed to create cache entries by hand, so, unless we want someone to do that, we shouldn't expose these functions.
> 
> What do you think?

Thanks for the quick reply!

Avoiding unneeded exports makes totally sense. But in this case, we only allow creating objects in a defined interface format. And i don't think somebody want's to create entries or responses by hand. Except testing.

I did't validate, but i promise the go compiler includes all methods of an imported package into a production build.
And exporting input and output objects of a package simplifies writing tests.
Keeping test-code out of a production-build and making it possible to write simple test has bigger priority - from my point of view.

danielpoe:
*_test files are not included in binary - also if they are not in _test package ( https://stackoverflow.com/questions/44685206/will-go-test-code-only-referenced-in-test-files-be-compiled-into-the-binary )  

In general I consider testing internal "units" that we dont want to make part of public "API" with non _test packages in some cases NOT as bad practice - if they are useful and easy to understand

jochen42:
ok. i just had a look at the compile-code and did not found any exclusion by package name. 
but there is a dead-code detection and elimination.

the compiles compiles the whole code of a package, detects unused code and removes it from the binary. so yes, the test-code will not be in the binary.

call me idiomatic ;) my personal preference is to make this explicit instead of let the compiler do this implicit. explicit is always easier to understand - for me.

i am not married with it, so i will close the pr.

bastianccm:
Just to clarify: the compiler cleans everything up, no need to discuss that point, but it's a programmatic API we would need to maintain (e.g.: exposing a function `func NewEntryMeta(lifetime time.Time, gracetime time.Time, tags []string) Meta {` puts a constraint on the code that any metadatas internal data can be constructed out of the "lifetime", "gracetime" and "tags". If we want to change that we would be constrained in the future.)

PR #87, opened 2019-11-10
Multilevel/Redis cache backend

jochen42:
A simple implementation of an MultiLevel-Cache-Backends and an Redis-Cache-Backend.

The redisBackend can be used as a second-level cache together with the inMemoryBackend to reduced the pressure to http-backends in case of large traffic-increase in a short time-period.

The RedisBackend has a simple Lock-Mechanism which prevents parallel writes of the same cache-key.

Another idea is to creata an awsS3 Backend. This will remove the need of manage an additional redis-instance. It is a little slower than redis, but scales really well.

Additional i created a common backend test case, which can tests the basic functionality of all implemented backends.

This is a first concept. Implementation to be discussed. I will add some documentation the next days.


core/cache/redisBackend.go
```diff
@@ -0,0 +1,373 @@
...
+		Password:    "",
+		MaxIdle:     8,
+		IdleTimeout: time.Minute * 30,
+		WriteLock:   false,
+	}
+}
+
+// NewRedisBackend creates an redis cache backend
+func NewRedisBackend(options RedisBackendOptions, logger flamingo.Logger) Backend {
+	mergo.Merge(&options, DefaultRedisBackendOptions())
```
golangcibot: Error return value of `mergo.Merge` is not checked (from `errcheck`)

core/cache/redisBackend.go
```diff
@@ -0,0 +1,373 @@
...
+		)
+		if err != nil {
+			b.logger.WithField("category", "redisBackend").Error("Error setting tag: %v: %v", tag, key)
+			return err
+		}
+	}
+
+	conn.Flush()
+	if b.writeLockEnabled {
+		b.unLock(conn, key)
```
golangcibot: Error return value of `b.unLock` is not checked (from `errcheck`)

core/cache/redisBackend_test.go
```diff
@@ -0,0 +1,78 @@
...
+		log.Fatalf("Could not connect to redis-docker: %s", err)
+	}
+	err = redis.Conn.Close(connection)
+	if err != nil {
+		log.Fatalf("Could not close redis-docker: %s", err)
+	}
+}
+
+// teardown the redis docker-container
+func teardown() {
```
golangcibot: `teardown` is unused (from `deadcode`)

core/cache/redisBackend.go
```diff
@@ -0,0 +1,378 @@
...
+	err := mergo.Merge(&options, DefaultRedisBackendOptions())
+	if err != nil {
+		logger.WithField("category", "redisBackend").Error(fmt.Sprintf("Error merging options: %v", err))
+	}
+
+	b := &RedisBackend{
+		Pool: &redis.Pool{
+			MaxIdle:     options.MaxIdle,
+			IdleTimeout: options.IdleTimeout,
+			TestOnBorrow: func(c redis.Conn, t time.Time) error {
```
danielpoe: nice

core/cache/redisBackend.go
```diff
@@ -0,0 +1,378 @@
...
+			},
+			Dial: func() (redis.Conn, error) {
+				return redisConnector(options.Network, fmt.Sprintf("%v:%v", options.Host, options.Port), options.Password)
+			},
+		},
+		logger: logger,
+	}
+	b.withBackendType("redis")
+
+	runtime.SetFinalizer(b, finalizer) // close all connections on destruction
```
danielpoe: just learned about that :-) - sounds useful

core/cache/redisBackend.go
```diff
@@ -0,0 +1,378 @@
...
+		if _, err := c.Do("AUTH", password); err != nil {
+			c.Close()
+			return nil, err
+		}
+	}
+	return c, err
+}
+
+func finalizer(b *RedisBackend) {
+	b.Close()
```
danielpoe: seems the only place that calls Close - can Close be private?

core/cache/redisBackend.go
```diff
@@ -0,0 +1,378 @@
...
+	b.logger = logger
+}
+
+// Close ensures all redis connections are closed
+func (b *RedisBackend) Close() {
+	b.Pool.Close()
+}
+
+// CreatePrefixedKey creates an redis-compatible key
+func (b *RedisBackend) CreatePrefixedKey(key string, prefix string) string {
```
danielpoe: private?

Also I think prefixing should include cachename - if multiple caches use same redis.

I think the whole "factory" logic of creating frontend with backend(s) and the corresponding "Hitcounter" initialisation makes sense to have a short brainstorming on - because there are some invariant that makes sense to enforce.

core/cache/redisBackend.go
```diff
@@ -0,0 +1,378 @@
...
+	}
+
+	b.countHit()
+	return b.buildResult(redisEntry), true
+}
+
+// Set an cache key
+func (b *RedisBackend) Set(key string, entry *Entry) error {
+	conn := b.Pool.Get()
+	defer conn.Close()
```
danielpoe: Can you explain why closing the connection?

Second thought: What if multiple cache frontends use backends with same redis - does it make sense to reuse connections?

core/cache/redisBackend.go
```diff
@@ -0,0 +1,378 @@
...
+
+	if b.writeLockEnabled {
+		err := b.lock(conn, key)
+		if err != nil {
```
danielpoe: unlock should be in a defer call rigth the next line

Early exists in case of errors below would not free the lock

core/cache/redisBackend.go
```diff
@@ -0,0 +1,378 @@
...
+import (
+	"bytes"
+	"encoding/gob"
+	"fmt"
+	"regexp"
+	"runtime"
+	"time"
+
+	"flamingo.me/flamingo/v3/framework/flamingo"
+	"github.com/gomodule/redigo/redis"
```
danielpoe: any reasons why this package?

I found also: https://github.com/go-redis/redis   which looks nice from the API

(Did not do a compare - just want to know)

core/cache/redisBackend.go
```diff
@@ -0,0 +1,378 @@
...
+		return err
+	}
+
+	conn.Flush()
+
+	return nil
+}
+
+func (b *RedisBackend) encodeEntry(entry *RedisCacheEntry) (*bytes.Buffer, error) {
+	gob.Register(entry)
```
danielpoe: shouldn't we expect that the caller has called this before?

core/cache/backendHitCounter.go
```diff
@@ -0,0 +1,46 @@
...
+
+var (
+	backendCacheKeyType, _ = tag.NewKey("backend_type")
+	backendCacheHitCount   = stats.Int64("flamingo/cache/backend/hit", "Count of cache-backend hits", stats.UnitDimensionless)
+	backendCacheMissCount  = stats.Int64("flamingo/cache/backend/miss", "Count of cache-backend misses", stats.UnitDimensionless)
+)
+
+type (
+	// BackendHitCounter representation
+	BackendHitCounter struct {
```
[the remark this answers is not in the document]
danielpoe: That might also be a usecase for https://golang.org/doc/go1.4#internalpackages ?

core/cache/redisBackend.go
```diff
@@ -0,0 +1,387 @@
...
+	}
+}
+
+// Lock tries to get an remote lock for an key
+func (b *RedisBackend) lock(conn redis.Conn, key string) (err error) {
+	reply, err := conn.Do(
+		"EXISTS",
+		b.CreatePrefixedKey(key, lockPrefix),
+	)
+	lockValue, err := redis.String(reply, err)
```
golangcibot: SA4006: this value of `err` is never used (from `staticcheck`)

core/cache/inMemoryBackend_test.go
```diff
@@ -0,0 +1,17 @@
...
+import (
+	"testing"
+
+	"flamingo.me/flamingo/v3/core/cache"
+)
+
+func Test_RunDefaultBackendTestCase_InMemoryBackend(t *testing.T) {
+	backend := cache.NewInMemoryCache()
+
+	testcase := &cache.BackendTestCase{
```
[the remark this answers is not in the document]
jochen42: > Like the idea with the Testing Lib.
> * Some go packages add this to a subpackage "testlibs" - I found no official recommendation
moving the the file to an sub-package would case an import cycle, because the backend interface is needed there
> * We can even add a Factory Method to avoid missuse:  NewBackendTest(t,backend)
makes sense i will change

core/cache/backendTestCase_test.go
```diff
@@ -0,0 +1,167 @@
...
+}
+
+type (
+	BackendTestCase struct {
+		T       *testing.T
+		Backend Backend
+	}
+)
+
+func (tc *BackendTestCase) RunTests(backendSaveTagsInMeta bool) {
```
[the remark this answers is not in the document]
jochen42: because in redis we are putting the tags to additional keys (SADD and SMEMBERS). if we would put the keys additionaly to the value of the key, we would have unessacary network-traffic and memory usage.

core/cache/redisBackend.go
jochen42: changed

core/cache/redisBackend.go
jochen42: because it is already used here: https://github.com/i-love-flamingo/flamingo/blob/master/core/healthcheck/domain/healthcheck/redisSession.go#L3

and i do not want to integrate a second library for the same thing...

core/cache/redisBackend.go
jochen42: do you mean the "gob.Register" call, or the whole encodeEntry?

core/cache/redisBackend.go
jochen42: as you can read [here](https://godoc.org/github.com/gomodule/redigo/redis#Pool), the conn.Close() doesn't close the connection itself, it just put's the connection to the pool. this mechanism allows you to reduce overhead of doing connection-initiations by keeping some connections alive. the pool is configurable via RedisBackendOptions.MaxIdle and RedisBackendOptions.IdleTimeout

core/cache/backendHitCounter.go
```diff
@@ -0,0 +1,46 @@
...
+		backendType string
```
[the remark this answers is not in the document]
jochen42: what do you expect as an value for the cacheName?

core/cache/cache.go
```diff
@@ -28,6 +28,8 @@ type (
 		Purge(key string) error
 		PurgeTags(tags []string) error
 		Flush() error
+		TagSupport() bool
```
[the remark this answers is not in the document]
jochen42: > Semantically you mean "PurgeByTagsSupport" - right?
i don't know why the PurgeTags was an part of the interface before, because no existing cache-backend implements it. but i see more usages for tags-support. f.e GetByTag (which can usage redis MGET-feature).
since it is not use currently, i would suggest to remove the PurgeTags completly.
For the Flush i would suggest to implement it in all backends (only file backend missed). should be a simple task.

> 
> Another option would be to make the two methods above Flush and PurgeTags not part of the main backend interface but add additional interfaces that we can type assert on the places where we want to check for them.
> We use this way of "adding" optional behavior to interfaces on other places as well
Sounds nice. i'll try.

core/cache/multiLevelBackend.go
```diff
@@ -0,0 +1,128 @@
...
+
+	return nil, false
+}
+
+// Set entry for key
+func (mb *MultiLevelBackend) Set(key string, entry *Entry) error {
+	for _, backend := range mb.backends {
+		err := backend.Set(key, entry)
+		if err != nil {
+			// We don't want fail here, only error-log
```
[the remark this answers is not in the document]
jochen42: good question. all current frontend implementation are ignoring this. don't know why:

https://github.com/i-love-flamingo/flamingo/blob/master/core/cache/httpFrontend.go#L177
https://github.com/i-love-flamingo/flamingo/blob/master/core/cache/stringFrontend.go#L58

i will return an error, if all backends fails.

core/cache/redisBackend.go
```diff
@@ -0,0 +1,378 @@
...
+	"github.com/gomodule/redigo/redis"
+	"github.com/imdario/mergo"
+	"github.com/pkg/errors"
+)
+
+type (
+	// RedisBackend instance representation
+	RedisBackend struct {
+		BackendHitCounter
+		Pool             *redis.Pool
```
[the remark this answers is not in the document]
jochen42: changed.

core/cache/fileBackend.go
```diff
@@ -28,9 +29,11 @@ func NewFileBackend(baseDir string) *FileBackend {
 		baseDir = defaultBaseDir
 	}
 
-	return &FileBackend{
+	fb := &FileBackend{
 		baseDir: baseDir,
 	}
+	fb.withBackendType("inMemory")
```
panzerfahrer: Shouldn't this be `fileBackend`?

core/cache/multiLevelBackend.go
```diff
@@ -0,0 +1,128 @@
...
+		logger   flamingo.Logger
+	}
+	// MultiLevelBackendOptions representation
+	MultiLevelBackendOptions struct {
+		Backends []Backend
+	}
+)
+
+// NewMultiLevelBackend creates a MultiLevelBackend isntance
+func NewMultiLevelBackend(options MultiLevelBackendOptions, logger flamingo.Logger) *MultiLevelBackend {
```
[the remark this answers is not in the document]
panzerfahrer: Isn't this redundant with the `Inject()` also setting a logger?

core/cache/fileBackend.go
jochen42: ;) yes. fixed

core/cache/multiLevelBackend.go
```diff
@@ -0,0 +1,128 @@
...
+func (mb *MultiLevelBackend) Inject(logger flamingo.Logger) {
+	mb.logger = logger
+}
+
+// Get entry by key
+func (mb *MultiLevelBackend) Get(key string) (entry *Entry, found bool) {
+	for _, backend := range mb.backends {
+		entry, found := backend.Get(key)
+		if found {
+			return entry, found
```
panzerfahrer: You set key/value on all backends but get it only from one. Isn't this potentially inconsistent?

jochen42: yes, that's the wanted behavior. both caches should always have the newest entry-version. but we take entry of the first cache which contains the key. in most cases this should be the first cache. only in scalup-cenarios the first cache will be empty and the second (mostly shared) cache will have the entry.

core/cache/redisBackend.go
bastianccm: Finalizers are not destructors or anything! They might not be called, from docs:

The finalizer is scheduled to run at some arbitrary time after the program can no longer reach the object to which obj points. There is no guarantee that finalizers will run before a program exits, so typically they are useful only for releasing non-memory resources associated with an object during a long-running program

bastianccm: Finalizers are not destructors or anything! They might not be called, from docs:

The finalizer is scheduled to run at some arbitrary time after the program can no longer reach the object to which obj points. There is no guarantee that finalizers will run before a program exits, so typically they are useful only for releasing non-memory resources associated with an object during a long-running program

panzerfahrer: How about using the `flamingo.ShutdownEvent` instead?

bastianccm: It was mere an informational info: closing the connection is only necessary when the flamingo stays connected, and/or to finish/flush data.
If the process ends, the connection is closed (by the operating system). We can use flamingo.ShutdownEvent to be sure that the connection is closed, or finalizers to "probably close" it if, by any chance, the cache backend is garbage collected (which actually shouldn't happen?).

So, I'm not against finalizers here, it's mere an info that this is not a destructor as e.g. in java/c++.

core/cache/redisBackend.go
bastianccm: It's enough to do `func init() { gob.Register(new(RedisCacheEntry)) }` once. gob only needs to look at the struct type, not the values.

Actually, calling gob.Register later, can lead to confusions/race conditions, such as decoding an entry via gob before it was initially registered if using interfaces inside.

core/cache/cache.go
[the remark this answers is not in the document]
bastianccm: +1 on type assertions. Do not change an existing, exposed interface (backwards incompatible)

core/cache/cache.go
```diff
@@ -28,6 +28,8 @@ type (
...
+		FlushSupport() bool
```
bastianccm: Same as above. Also I would not add this at all. Just call `Flush`. If the backend doesn't support it, the backend can ignore it. The interface wants a `Flush` method which makes sense for caches, so we should leave this to the interface and not the users of the interface to decide what happens.

```
func (*B) FlushSupport() bool { return false }
func (*B) Flush() { }
// ...
if backend.FlushSupport() {
    backend.Flush()
}
```
is pretty much the same as
```
func (*B) Flush() { }
// ...
backend.Flush()
```
however in the later case the compiler can optimize and we omit one function call and branch.

danielpoe:
> Can we split this PR and move the general cache stuff (Metrics etc) into a new PR? This one can stay for the redis stuff I think?
> This one gets too big for a proper review.

I wouldn't split this now - not worth the extra effort separating everything again.
Lets do proper QA/ Review on that one and finish the feature completly.
The complete cache modul benefits from it and was in a very initial state anyhow

core/cache/fileBackend.go
```diff
@@ -28,9 +29,14 @@ func NewFileBackend(baseDir string) *FileBackend {
 		baseDir = defaultBaseDir
 	}
 
-	return &FileBackend{
-		baseDir: baseDir,
+	os.MkdirAll(baseDir, os.ModePerm)
```
golangcibot: Error return value of `os.MkdirAll` is not checked (from `errcheck`)

core/cache/multiLevelBackend.go
```diff
@@ -0,0 +1,99 @@
...
+	for _, backend := range mb.backends {
+		err := backend.Purge(key)
+		if err != nil {
+			errorList = append(errorList, err)
+			mb.logger.WithField("category", "multiLevelBackend").Error(fmt.Sprintf("Failed Purge with error %v", err))
+		}
+	}
+
+	if 0 != len(errorList) {
+		return errors.New(fmt.Sprintf("Not all backends succeeded to Purge key %v, Errors: %v", key, errorList))
```
golangcibot: S1028: should use fmt.Errorf(...) instead of errors.New(fmt.Sprintf(...)) (from `gosimple`)

core/cache/multiLevelBackend.go
```diff
@@ -0,0 +1,99 @@
...
+	for _, backend := range mb.backends {
+		err := backend.Flush()
+		if err != nil {
+			errorList = append(errorList, err)
+			mb.logger.WithField("category", "multiLevelBackend").Error(fmt.Sprintf("Failed Flush error %v", err))
+		}
+	}
+
+	if 0 != len(errorList) {
+		return errors.New(fmt.Sprintf("Not all backends succeeded to Flush. Errors: %v", errorList))
```
golangcibot: S1028: should use fmt.Errorf(...) instead of errors.New(fmt.Sprintf(...)) (from `gosimple`)

core/cache/cache.go
jochen42: ok. i change the interface Backend to it's original version and a added a new interface TagSupportingBackend.

But this has currently one caveat: The multilevel backend is not able to support tags. for the moment this is ok. we could allow to add TagSupportingBackends into MultiLevelBackendOptions to solve this. But i don't think this would be a future proof-solution, because we don't know if we would have another type of backend-features.
And we could allow to add an interface{} to MultiLevelBackendOptions, but this would break the type-safety.

please let me now, if you have an prefered option here.

core/cache/multiLevelBackend.go
golangcibot: S1028: should use fmt.Errorf(...) instead of errors.New(fmt.Sprintf(...)) (from `gosimple`)

core/cache/multiLevelBackend.go
golangcibot: S1028: should use fmt.Errorf(...) instead of errors.New(fmt.Sprintf(...)) (from `gosimple`)

core/cache/redisBackend.go
jochen42: nice. thanks for the hint!

jochen42:
@danielpoe we talked about an tags "frontend-name" for the cache-backend metrics.

i did a small test-implementation like we both commited.
(Add an parameter at backend instance creation)
But this will cause an interface change of all current cache-usages.
The existing backend-factories (file, inMemory) would need this parameter and since go doesn't support default paramater values, i currently don't have a solution to do this backward-compatible.

an alternative way would be to change the interface of the "backend" interface. (add parameter frontend-name). this would help to make a backend-instance reusable for multiple frontend's and it would not be a problem for existing projects, since the backend-interface is only used by httpFrontend and stringFrontend. But yes. also not nice. ;)

any ideas?

cc @bastianccm

core/cache/Readme.md
```diff
@@ -56,6 +56,32 @@ response, err := apiclient.Cache.Get(requestContext, u.String(), loadData)
 ## Cache backends
 
 Currently there are the following backends available:
-* inMemoryCache (caches in memory - and therefore is a very fast cache)
-* fileBackend (caches in filesystem )
-* nullBackend (caches nothing)
\ No newline at end of file
+
+### inMemoryCache
```
jochen42: @danielpoe, @bastianccm 

I started some basic documentation and i am open for input.

core/cache/redisBackend.go
jochen42: ah ok. i only added the finalizer here, to ensure the connections are closed.
the reason was:
After adding the docker-tests, the docker containers were not destroyed, because of open connections. after adding the finalizer the connections there closed correctly and the containers are gone ofter the tests are done.

as you described, in production it does not matter. But currently i don't have an other solution, expect adding and exported "Close"-method to the RedisCacheBackend. But exporting an Method only for teardown an test - i am not sure ;)

core/cache/multiLevelBackend.go
golangcibot: S1028: should use fmt.Errorf(...) instead of errors.New(fmt.Sprintf(...)) (from `gosimple`)

core/cache/multiLevelBackend.go
golangcibot: S1028: should use fmt.Errorf(...) instead of errors.New(fmt.Sprintf(...)) (from `gosimple`)

jochen42:
@danielpoe @bastianccm 

it seems, that the httpFrontend caches also error-responses. this is dangerous with an shared cache-backend. i would suggest to extend the frontend to cache only 4xx-error response, but not 5xx.

what do you think?

danielpoe:
> @danielpoe @bastianccm
> 
> it seems, that the httpFrontend caches also error-responses. this is dangerous with an shared cache-backend. i would suggest to extend the frontend to cache only 4xx-error response, but not 5xx.
> 
> what do you think?

Thats on purpose - if the backend responses with an error you dont need to retry.
But the cache time is much shorter..
Its a kind of "circuit breaker"

jochen42:
> > @danielpoe @bastianccm
> > it seems, that the httpFrontend caches also error-responses. this is dangerous with an shared cache-backend. i would suggest to extend the frontend to cache only 4xx-error response, but not 5xx.
> > what do you think?
> 
> Thats on purpose - if the backend responses with an error you dont need to retry.
> But the cache time is much shorter..
> Its a kind of "circuit breaker"

ok. i am not a fan of this, because in kubernetes we would break the circuit to a service-address not a pod.  if the backend-service has multiple pods, no retry on an other backend-pod will be done. and with the shared cache, the error will be cached in all flamingo-pods.

the circuit should be done at another place. perhaps we can make this frontend-behavior configurable?

danielpoe:
Comments from Meeting 4.12.2019:

1) Multilevel Backend -> Adjust to TwoLevelBackend with a Slow and a Fast backend. 
 Misses in Fast and Hit in Slow should update Fast.

2) Remove useless current Locking in Redis Backend

3) Simple CacheFactory stuff

Addon Required: Think about a shared Lock: Because now cache entries expired still lead to thundering because all flamingo pods will try to refetch expired cache entries
=> New PR

core/cache/redisBackend.go
```diff
@@ -0,0 +1,331 @@
...
+	"github.com/imdario/mergo"
+)
+
+type (
+	// RedisBackend instance representation
+	RedisBackend struct {
+		CacheMetrics     CacheMetrics
+		Pool             *redis.Pool
+		logger           flamingo.Logger
+		writeLockEnabled bool
```
golangcibot: `writeLockEnabled` is unused (from `structcheck`)

core/cache/redisBackend.go
```diff
@@ -0,0 +1,331 @@
...
+		MaxIdle int
+		// Timout to close idle connections (default 5m)
+		IdleTimeout time.Duration
+		// WriteLock feature switch
+		WriteLock bool
+	}
+)
+
+const (
+	lockPrefix  = "lock:"
```
golangcibot: `lockPrefix` is unused (from `varcheck`)

core/cache/twoLevelBackend.go
```diff
@@ -0,0 +1,116 @@
...
+// Get entry by key
+func (mb *TwoLevelBackend) Get(key string) (entry *Entry, found bool) {
+	entry, found = mb.firstBackend.Get(key)
+	if found {
+		return entry, found
+	}
+
+	entry, found = mb.secondBackend.Get(key)
+	if found {
+		go mb.firstBackend.Set(key, entry)
```
golangcibot: Error return value of `mb.firstBackend.Set` is not checked (from `errcheck`)

core/cache/redisBackend.go
```diff
@@ -0,0 +1,269 @@
...
+
+	// RedisCacheEntry representation
+	RedisCacheEntry struct {
+		Meta redisCacheEntryMeta
+		Data interface{}
+	}
+)
+
+const (
+	lockPrefix  = "lock:"
```
golangcibot: `lockPrefix` is unused (from `varcheck`)

jochen42:
@danielpoe @bastianccm 

i did some rework we talked about. open point's are

- CacheFactory (i only heard the last minute of your plans. either write down your plan here or somebody else can implement)
- handling frontendName using dingo
- documentation

core/cache/factory_test.go
```diff
@@ -0,0 +1,73 @@
...
+		"two": config2.Map{
+			"backendType": "inmemory",
+			"inMemoryBackend": config2.Map{
+				"size": 100.0,
+			},
+		},
+	}
+
+	var typedCacheConfig FactoryConfig
+	testconfig.MapInto(&typedCacheConfig)
```
golangcibot: Error return value of `testconfig.MapInto` is not checked (from `errcheck`)

danielpoe:
closing in favor of https://github.com/i-love-flamingo/flamingo/pull/95/files

PR #89, opened 2019-11-19
Add config to disable OIDC

danielpoe:
Oauth and authmanager might be an dependency for other modules (e.g. cart module)
But if "login" is never used for a certain project the current authmanager will always pani of no valid OIDC provider is configured.

So this PR suggest to allow to disable the oidc feature by a flag.

Its a intermediate solution until the major new auth concept is included.


core/oauth/application/authmanager.go
```diff
@@ -78,7 +78,7 @@ func (f *loggingRoundTripper) RoundTrip(req *http.Request) (*http.Response, erro
 }
 
 // Inject authManager dependencies
-func (am *AuthManager) Inject(logger flamingo.Logger, router *web.Router, openIDProvider *oidc.Provider, config *struct {
```
bastianccm: This change broke backwards compatibility between v3.0.3 and v3.0.4

PR #93, opened 2019-12-03
Add caching for the TranslationService api

AQUI74S:


danielpoe:
What is the performance improvement?
Does the underlying implementation does also cache the loaded translationfiles - so not sure what is the benefit here?

bastianccm:
It's the amount of data which is super expensive (takes 2s for some projects), that's why we cache the whole generated list here

danielpoe:
> It's the amount of data which is super expensive (takes 2s for some projects), that's why we cache the whole generated list here

Sure - also in non devMode?   (devMode cache for loading translationfiles is deactive)

PR #98, opened 2019-12-18
Globallock and Global Single Process proposal

danielpoe:


core/cluster/process.go
```diff
@@ -0,0 +1,38 @@
...
+type (
+
+	SingleProcess interface {
+		//Do runs the given function if there is no other function with that key running
+		Do(key string, fn func()) (bool, error)
+	}
+
+	DistributedLock interface {
+		//GetLock - call to requests a distributed lock - if not possible. returns "false" in case Lock is not available without blocking
+		GetLock(key string, max *time.Duration) (bool, error)
```
bastianccm: Use `context.Context` instead of `time.Duration` to react on canceled contexts. Timeout can be achieved by using `context.WithTimeout` or `context.WithDeadline`.

bastianccm: Rename to `Lock`.

core/cluster/process.go
```diff
@@ -0,0 +1,38 @@
...
+		Unlock(key string)
+	}
+
+	SingleProcessImpl struct {
+		lock DistributedLock
+	}
+)
+
+//Do runs the func if Lock can be obtained
+func (s *SingleProcessImpl) Do(key string, fn func()) (bool, error) {
```
bastianccm: Unclear what a result of `false, nil` means: not run but no error? Then `Do` would be the wrong naming, as it might or might not run the function.

core/cluster/process.go
```diff
@@ -0,0 +1,38 @@
...
+import "time"
+
+type (
+
+	SingleProcess interface {
+		//Do runs the given function if there is no other function with that key running
+		Do(key string, fn func()) (bool, error)
+	}
+
+	DistributedLock interface {
```
bastianccm: Should with be compatible with `sync.Locker`?

core/cluster/process.go
```diff
@@ -0,0 +1,38 @@
+package cluster
+
+import "time"
+
+type (
+
+	SingleProcess interface {
```
bastianccm: Rename to `Once`?

core/cluster/process.go
```diff
@@ -0,0 +1,38 @@
+package cluster
```
bastianccm: We could name the package `clustersync` and provide semantics similar to the `sync` package.
`sync.Once` for running once.
`sync.Mutex` for a mutual-exclusive lock.

core/cluster/process.go
```diff
@@ -0,0 +1,38 @@
...
+	SingleProcess interface {
+		//Do runs the given function if there is no other function with that key running
+		Do(key string, fn func()) (bool, error)
+	}
+
+	DistributedLock interface {
+		//GetLock - call to requests a distributed lock - if not possible. returns "false" in case Lock is not available without blocking
+		GetLock(key string, max *time.Duration) (bool, error)
+		//Free the distributed Lock
+		Unlock(key string)
```
bastianccm: Can `Unlock` fail? As we do rely on network operation for distribution handling we might want to keep that into account?

core/cluster/process.go
danielpoe: Once and Locker works because of using same instance of the type. Which is replaced by the key paramter. so it cannot be interface compatible but the renaming makes sense

core/cluster/process.go
danielpoe: Lock blocks
this function should nit block
maybe there is another better name..

danielpoe:
closing since we decided to add the feature first local where it is required - see flamingo-commerce cart package

PR #102, opened 2020-01-07
Extract app init

danielpoe:
For better black box testing of an Flamingo App its currently hard to run an App with the core logic - since everything happens inside App() method 

If we split the logic its possible to get an fully initialized area (that can be used to get initialised injector etc)


bastianccm:
I pushed my proposed extensions to that concept

PR #108, opened 2020-01-23
Adding EmptySessionWithID to web.session for testing purposes

masmrlar:
The additional function allows to create sessions for tests with a set ID.
The ID is usually set by the session store, but as soon as a test requires sessions with ID things start to get complicated.

### DIFF
--- a/framework/web/session.go
+++ b/framework/web/session.go
@@ -23,6 +23,15 @@ func EmptySession() *Session {
 	return &Session{s: sessions.NewSession(nil, "")}
 }
 
+// EmptySessionWithID creates an empty session instance with a preset session ID for testing etc.
+func EmptySessionWithID(id string) *Session {
+	result := EmptySession()
+
+	result.s.ID = id
+
+	return result
+}
+
 // ContextWithSession returns a new Context with an attached session
 func ContextWithSession(ctx context.Context, session *Session) context.Context {
 	return context.WithValue(ctx, contextSession, session)


danielpoe:
> The additional function allows to create sessions for tests with a set ID.
> The ID is usually set by the session store, but as soon as a test requires sessions with ID things start to get complicated.

Can you add a example why a test would need to force an id?

masmrlar:
@danielpoe it's a test on a func using IDHash() of web.Session. 

The gorilla session is hidden in web.Session, but we have web.Session within the context as parameter everywhere. 

The test wants to check if different users somewhere really get a different hash code, where the algorithm also takes IDHash() into account. 

So the easiest solution is to have prepared sessions for tests with a set ID - setting the ID on the session is usually only done by the session storage. We don't want to mock the whole request/create session from storage stack to get the same result (where we are still not able to get a web.Session containing a session with ID as imho this is all hidden in the flamingo.web package.

danielpoe:
> @danielpoe it's a test on a func using IDHash() of web.Session.
> 
> The gorilla session is hidden in web.Session, but we have web.Session within the context as parameter everywhere.
> 
> The test wants to check if different users somewhere really get a different hash code, where the algorithm also takes IDHash() into account.
> 
> So the easiest solution is to have prepared sessions for tests with a set ID - setting the ID on the session is usually only done by the session storage. We don't want to mock the whole request/create session from storage stack to get the same result (where we are still not able to get a web.Session containing a session with ID as imho this is all hidden in the flamingo.web package.

Sorry still don't understand really :-(
I am asking because adding methods just for test is always not perfect.

If the test just "reads" the sessionid and want to make sure that another "read" is not the same that could be done without seeting an id before? maybe still misunderstood by me..

masmrlar:
Ok - i try to explain (but I also agree - adding public methods just for testing is not a really good practice).

We 'got a function calculating a hash code per request and must ensure that the code differs per session where each request has the same path/parameters. 

So we want to use the IDHash() of Session and take that into account. 

Problem ist now that IDHash() requires to have an ID on the session - so our test fails by just using EmptySession(). 

I'll try to find another way to get this result without the additional testing method - but it could bloat the rather slim test code as we would need to use web.handler.ServeHTTP somehow as the session is handled in getSession() in flamingo/framework/web/handler.go.

Another possibility could be to just store a random value within the session to gain the same result. But this also feels uncomfortable as the session ID should be the thing distinguishing sessions, not a helper value stored in the session (imho).

masmrlar:
@danielpoe ok - we can drop it. 
I found a way to test it without adding this function. Took me a while, but I also got some more insights on how the routing and request handling works :)

PR #110, opened 2020-01-30
add base path to ServerErrorResponse

doertydoerk:
The root path in my project is configurable via
flamingo.router.path: %%ENV:BASE_PATH%%/gateway/%%

My error pages are static HTML. So, if the above path changes all CSS and assets of my error pages won't be found anymore.
Having base available in the DataResponse would allow me to set the path "dynamically" in my static HMTL like so: <base href="{{ .base }}/">

Considering all other options and given this is an edge case for a Flamingo environment, I find this to be the least invasive option to the problem.

Thanks for approving my request!


doertydoerk:
screwed up the full request, will submit a new
 one

PR #112, opened 2020-01-31
add base path to ServerErrorResponse

doertydoerk:


bastianccm:
Why don't you use the `url` template function? It will handle the base path correctly already if the path starts with `/`.
https://github.com/i-love-flamingo/flamingo/blob/master/framework/web/router.go#L168

bastianccm:
Also for the mocks we don't have a `go:generate` to regenerate the mocks, like here: https://github.com/i-love-flamingo/flamingo/blob/master/core/security/application/doc.go#L3

Are the mocks necessary anyway?

And please sort the imports in `framework/web/result_test.go`

bastianccm:
Resolved by using `url "/"` template function. This will be reverted.

PR #120, opened 2020-02-13
core/locale: module tests and refactorings

Ompluscator:
1. missing unit tests
2. fixed file names
3. fixed naming inconsistencies


core/locale/interfaces/templatefunctions/price.go
```diff
@@ -8,6 +8,25 @@ import (
 	"flamingo.me/flamingo/v3/framework/config"
 )
 
+// PriceFormatFunc for formatting prices
+type PriceFormatFunc struct {
+	config       config.Map
```
golangcibot: `config` is unused (from `structcheck`)

core/locale/interfaces/templatefunctions/label_test.go
```diff
@@ -0,0 +1,47 @@
+package templatefunctions
+
+import (
```
bastianccm: Please resort the imports

core/locale/domain/label.go
```diff
@@ -78,27 +70,27 @@ func (l *Label) SetCount(count int) *Label {
 	return l
 }
 
-// SetLocale on a label
-func (l *Label) SetLocale(localeCode string) *Label {
+// SetLocaleCode on a label
```
bastianccm: This is backwards incompatible (it changes the method names).
I would be fine with doing this, but it will require extra work for users of the locale package.

core/locale/application/date_time_service.go
```diff
@@ -8,22 +8,20 @@ import (
...
-	}
-)
+// DateTimeService is a basic support service for date/time parsing
+type DateTimeService struct {
+	dateFormat     string
+	timeFormat     string
+	dateTimeFormat string
+	location       string
+	logger         flamingo.Logger
+}
```
danielpoe: What is benefit of not grouping type definitions inside type () ?

core/locale/application/date_time_service_test.go
```diff
@@ -0,0 +1,133 @@
...
+)
+
+func TestDateTimeService_GetDateTimeFormatterFromIsoString(t *testing.T) {
+	start := time.Now()
+	now, err := time.Parse(time.RFC3339, start.Format(time.RFC3339))
+	assert.NoError(t, err, "no error received")
+
+	// error for invalid time
+	formatter, err := new(DateTimeService).GetDateTimeFormatterFromIsoString("invalid time")
+	assert.Nil(t, formatter, "got nil for formatter")
```
danielpoe: The third parameter is the error message. So better "expected nil for invalid time" is a better mesage.
Same is with every message in the assert.* usage

core/locale/application/date_time_service_test.go
```diff
@@ -0,0 +1,133 @@
...
+	dateTimeService.Inject(flamingo.NullLogger{}, &struct {
+		DateFormat     string `inject:"config:core.locale.date.dateFormat"`
+		TimeFormat     string `inject:"config:core.locale.date.timeFormat"`
+		DateTimeFormat string `inject:"config:core.locale.date.dateTimeFormat"`
+		Location       string `inject:"config:core.locale.date.location"`
+	}{
+		Location: "invalid location",
+	})
+
+	formatter, err = dateTimeService.GetDateTimeFormatterFromIsoString(now.Format(time.RFC3339))
```
danielpoe: Please do seperated subtest for each use case that improves readability of the test

core/locale/application/date_time_service_test.go
```diff
@@ -0,0 +1,133 @@
...
+		DateFormat:     time.RFC822,
+		TimeFormat:     time.Kitchen,
+		DateTimeFormat: time.ANSIC,
+		Location:       "America/New_York",
+	})
+
+	loc, err := time.LoadLocation("America/New_York")
+	assert.NoError(t, err, "no error received")
+
+	result := &domain.DateTimeFormatter{
```
danielpoe: This is expected result - should be named like this.

In general I am not sure if we should test it that way (knowing how the method constructs the formatter and rebuild this in the test) in general - maybe we can test it more blackbox:
calling formatter.FormatDateTime and compare the string with what we expect?

core/locale/application/price_service.go
```diff
@@ -0,0 +1,77 @@
+package application
```
danielpoe: Please use rename - so that in history the file is not deleted and recreated

core/locale/domain/date_time_test.go
```diff
@@ -0,0 +1,155 @@
+package domain
```
danielpoe: Same here - rename instead recreate

core/locale/domain/date_time_test.go
```diff
@@ -0,0 +1,155 @@
...
+
+func TestDateTimeFormatter_SetLocation(t *testing.T) {
+	now := getUTCNow()
+
+	formatter := &DateTimeFormatter{
+		logger:   flamingo.NullLogger{},
+		dateTime: now,
+	}
+
+	assert.Errorf(t, formatter.SetLocation("wrong"), ErrInvalidLocation)
```
danielpoe: Instead usage of ErrInvalidLocation the message should be "Expected Error on invalid location" - or better use assert.EqualError ?

core/locale/domain/date_time_test.go
```diff
@@ -0,0 +1,155 @@
...
+
+	loc, err := time.LoadLocation("Europe/Berlin")
+	assert.NoError(t, err)
+	assert.NoError(t, formatter.SetLocation(loc.String()))
+	assert.Equal(t, &DateTimeFormatter{
+		logger:        flamingo.NullLogger{},
+		dateTime:      now,
+		localDateTime: now.In(loc),
+	}, formatter)
```
danielpoe: For that block I think we should only use that line:

assert.NoError(t, formatter.SetLocation(loc.String()))


everything else is not testing this function

core/locale/domain/date_time_test.go
```diff
@@ -0,0 +1,155 @@
...
+}
+
+func TestDateTimeFormatter_SetLogger(t *testing.T) {
+	formatter := &DateTimeFormatter{}
```
danielpoe: Testing getter and setter in general is testing trivial things

My opinion is not to test this

core/locale/domain/label.go
danielpoe: Maybe we use this already in templates - so I would not break backward here

core/locale/application/date_time_service.go
Ompluscator: In general, if we want to use locale module to expose it with swagger spec, most of the libraries have an issue to fetch types defined in brackets pairs (like go-swagger). Second reason would be to have it consistent, as in most of the cases we use it without brackets. :)

core/locale/domain/label.go
Ompluscator: Added both, SetLocale is deprecated :)

core/locale/application/date_time_service.go
```diff
@@ -83,7 +85,7 @@ func (dts *DateTimeService) GetDateTimeFormatter(timeValue time.Time) (*domain.D
 
 func (dts *DateTimeService) loadLocation() (*time.Location, error) {
 	if dts.location == "" {
-		return nil, errors.New("no location configured")
+		return time.UTC, nil
```
carstendietrich, line 88: The whole if block is unneeded, `time.LoadLocation("")` will return UTC anyways.

core/locale/application/date_time_service.go
```diff
@@ -61,9 +63,9 @@ func (dts *DateTimeService) GetDateTimeFormatter(timeValue time.Time) (*domain.D
 	loc, err := dts.loadLocation()
 	if err != nil {
 		if dts.logger != nil {
-			dts.logger.Warn("dateTime Parsing error - could not load location - use UTC as fallback", dts.location)
+			dts.logger.Error("dateTime Parsing error - could not load location - use UTC as fallback", dts.location)
```
carstendietrich, line 66: The UTC fallback was removed please update log message accordingly.

carstendietrich (changes requested):
Hey @Ompluscator ,

sorry for the long waiting time on this one. Had a second look and found some breaking changes that we need to tackle before merging.

Some of the changes in the `DateTimeService` need some discussion with the rest of the maintainers, will give you feedback on that ASAP.

Feel free to fix the other stuff in the meantime 👌🏻

core/locale/interfaces/templatefunctions/label.go
```diff
@@ -9,22 +9,20 @@ import (
...
-)
+// LabelFormat is exported as a template function
+type LabelFormat struct {
+	labelService *application.LabelService
+	logger       flamingo.Logger
+}
 
 // Inject dependencies
-func (tf *Label) Inject(labelService *application.LabelService, logger flamingo.Logger) {
+func (tf *LabelFormat) Inject(labelService *application.LabelService, logger flamingo.Logger) {
```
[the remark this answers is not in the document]
Ompluscator: Actually, "Format" was dummy copy paste - I wanted to use LabelFunc :)

carstendietrich (approved):
@Ompluscator pushed some minor changes, will merge as soon as the pipeline is green.

PR #121, opened 2020-02-13
framework/config: remove obsolete note for number config injection

tessig:
integer injection is possible as well now.


bastianccm:
I find the example misleading. It is still possible to use float64 (and will always be). int is only possible if it is an actual integer. 
https://github.com/i-love-flamingo/flamingo/blob/master/framework/config/area.go#L332
Maybe we should state this more clearly?

PR #139, opened 2020-03-16
locale: add force file reload and add translation functions

jokaio:
force file reload: translation files are reloaded even if they are marked as loaded
add translation: a wrapper for the AddTranslation function of the i18n bundle

### DIFF
--- a/core/locale/infrastructure/translationService.go
+++ b/core/locale/infrastructure/translationService.go
@@ -4,6 +4,8 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	"github.com/nicksnyder/go-i18n/i18n/language"
+	"github.com/nicksnyder/go-i18n/i18n/translation"
 	"text/template"
 
 	"flamingo.me/flamingo/v3/core/locale/domain"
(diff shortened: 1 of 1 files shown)


danielpoe:
> force file reload: translation files are reloaded even if they are marked as loaded
> add translation: a wrapper for the AddTranslation function of the i18n bundle

Why and where is this needed?
I remember idea was to reload files if debug mode is on.

core/locale/infrastructure/translationService.go
```diff
@@ -4,6 +4,8 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	"github.com/nicksnyder/go-i18n/i18n/language"
```
bastianccm: Please run `goimports` to fix the import ordering

bastianccm:
@danielpoe this is to allow multiple translations/translation-sources and let other projects add and remove custom translations on the fly

danielpoe:
> @danielpoe this is to allow multiple translations/translation-sources and let other projects add and remove custom translations on the fly

Still dont get the use case by the descriptions in this PR.

My thoughts:
So an external package uses the interface layer directly? From which layer?
Methods like ForceFileReload are very implementation specific.

Why not implement your own TranslationService Adapter in the project (that may aggregate the default one)

PR #141, opened 2020-03-24
systemendpoint: do not panic on regular shutdown

tessig:
ListenAndServe returns ErrServerClosed on regular shutdown
which results in a panic here on each flamingo app shutdown

### DIFF
--- a/framework/systemendpoint/application/server.go
+++ b/framework/systemendpoint/application/server.go
@@ -56,7 +56,7 @@ func (s *SystemServer) Start() {
 	s.server = &http.Server{Addr: s.serviceAddress, Handler: serveMux}
 	go func() {
 		err := s.server.ListenAndServe()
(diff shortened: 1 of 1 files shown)


framework/systemendpoint/application/server.go
```diff
@@ -56,7 +56,7 @@ func (s *SystemServer) Start() {
 	s.server = &http.Server{Addr: s.serviceAddress, Handler: serveMux}
 	go func() {
 		err := s.server.ListenAndServe()
-		if err != nil {
+		if err != nil && err != http.ErrServerClosed {
```
bastianccm: Can you do
```golang
if err != nil && !errors.Is(err, http.ErrServerClosed)
```
to check for wrapped errors?

tessig: changed it

PR #143, opened 2020-04-30
auth.WebIdentityService.IdentifyAs

bastianccm:


panzerfahrer:
Is all this reflection really needed? How about introduce a new interface that visits the provider or identity returned by the provider?

PR #148, opened 2020-06-12
remove/depreacte whitelist/blacklist

bastianccm:


danielpoe:
Can we take the opportunity and document the related feature?

PR #158, opened 2020-07-27
move healthchecks to correct modules

### LINKED ISSUE #151: Decouple healthcheck from other modules
At the moment, the healthcheck module has a hard dependency on the session backend, but does not declare a dependency to the session module.

A similar situation is given for the auth check, but here the default setting is at least deactivated.

The healthcheck module should be usable in projects without sessions or auth, too.

I sugesst to move specific checks to their corresponding modules.

bastianccm:
Resolves #151

### DIFF
--- a/core/oauth/module.go
+++ b/core/oauth/module.go
@@ -3,8 +3,10 @@ package oauth
 import (
 	"flamingo.me/dingo"
 	"flamingo.me/flamingo/v3/core/auth"
+	"flamingo.me/flamingo/v3/core/healthcheck/domain/healthcheck"
 	"flamingo.me/flamingo/v3/core/oauth/application"
 	fakeService "flamingo.me/flamingo/v3/core/oauth/application/fake"
+	"flamingo.me/flamingo/v3/core/oauth/infrastructure"
 	"flamingo.me/flamingo/v3/core/oauth/interfaces"
 	fakeController "flamingo.me/flamingo/v3/core/oauth/interfaces/fake"
 	"flamingo.me/flamingo/v3/core/security/application/role"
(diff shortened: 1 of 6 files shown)


core/oauth/module.go
```diff
@@ -19,6 +21,7 @@ type Module struct {
 	UseFake                     bool   `inject:"config:core.oauth.useFake"`
 	PreventSimultaneousSessions bool   `inject:"config:core.oauth.preventSimultaneousSessions"`
 	SessionBackend              string `inject:"config:flamingo.session.backend"`
+	CheckAuthServer             bool   `inject:"config:core.healthcheck.checkAuth,optional"`
```
tessig, line 24: should we change the config path too? (and introduce a legacy config alias)

framework/flamingo/sessions.go
```diff
@@ -5,6 +5,8 @@ import (
 	"os"
 
 	"flamingo.me/dingo"
+	"flamingo.me/flamingo/v3/core/healthcheck/domain/healthcheck"
+	healthcheck2 "flamingo.me/flamingo/v3/framework/flamingo/healthcheck"
```
tessig, line 9: would choose a nice alias here

framework/flamingo/sessions.go
```diff
@@ -42,6 +45,7 @@ func (m *SessionModule) Inject(config *struct {
 	RedisPassword        string  `inject:"config:flamingo.session.redis.password"`
 	RedisIdleConnections float64 `inject:"config:flamingo.session.redis.idle.connections"`
 	RedisMaxAge          float64 `inject:"config:flamingo.session.redis.maxAge"`
+	CheckSession         bool    `inject:"config:core.healthcheck.checkSession,optional"`
```
tessig, line 48: same question for this config path

PR #163, opened 2020-08-16
framework/config: add printError before panic

danielpoe:
In case the config parsing fails flamingo panics.

Would be helpful to see the context (print the error lines)

### DIFF
--- a/framework/config/loader.go
+++ b/framework/config/loader.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"path/filepath"
 	"regexp"
+	"strconv"
 	"strings"
 
 	"cuelang.org/go/cue/format"
@@ -203,6 +204,7 @@ func loadYamlConfig(area *Area, config []byte) error {
 
 	cfg := make(Map)
 	if err := yaml.Unmarshal(config, &cfg); err != nil {
+		log.Print(errorLineDebug(err, config))
 		panic(err)
 	}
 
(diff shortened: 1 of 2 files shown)


framework/config/loader.go
```diff
@@ -213,6 +215,37 @@ func loadYamlConfig(area *Area, config []byte) error {
 	return area.loadedConfig.Add(cfg)
 }
 
+//errorLineDebug returns the lines where the error occurred (if possible)
+func errorLineDebug(err error, config []byte) string {
+	r, _ := regexp.Compile(": line (.*):")
```
SchiffFlieger, line 220: Only compile the regex once and use `MustCompile` to avoid errors.

framework/config/loader.go
```diff
@@ -213,6 +215,37 @@ func loadYamlConfig(area *Area, config []byte) error {
...
+	last := line + 10
+	if first < 0 {
+		first = 0
+	}
+	if last > len(lines) {
+		last = len(lines)
+	}
+	for i := first; i < last; i++ {
+		if i == line {
+			errorLines = errorLines + fmt.Sprintln(">", i, lines[i])
```
SchiffFlieger, line 241: The line numbers are off by 1 (zero-based counting).

SchiffFlieger (changes requested):

SchiffFlieger:
I like the way the error is presented, with the additional context it is very easy to spot which line caused the error. Unfortunately, we don't tell in which file the error occurred. Maybe it is better to not panic immediately, but instead bubble up the error and panic at a place where we can also tell the filename.

While playing around with the code, I realized that `github.com/ghodss/yaml` ignores some common errors in yaml files, e. g. duplicate keys. Not sure if this is wanted behavior. But that's something we should discuss separately.

SchiffFlieger (approved):

PR #167, opened 2020-10-07
legacy patch oauth

bastianccm:
do not fail if no new idtoken was provided (i.e. only access token was
updated)


jeinfeldt (approved):
A dream!

PR #171, opened 2020-11-01
framework/web: add handler as part of request

Ompluscator:
This code adds handler name attribute to web.Request and so possibility to have information if there is assigned handler and it's name later in ChainFilter, Middleware or Controller.

### DIFF
--- a/framework/web/handler.go
+++ b/framework/web/handler.go
@@ -112,7 +112,9 @@ func (h *handler) ServeHTTP(rw http.ResponseWriter, httpRequest *http.Request) {
 	_, span = trace.StartSpan(ctx, "router/matchRequest")
 	controller, params, handler := h.routerRegistry.matchRequest(httpRequest)
 
+	var handlerName string
 	if handler != nil {
+		handlerName = handler.handler
 		ctx, _ = tag.New(ctx, tag.Upsert(ControllerKey, handler.GetHandlerName()), tag.Insert(opencensus.KeyArea, "-"))
 		httpRequest = httpRequest.WithContext(ctx)
 		start := time.Now()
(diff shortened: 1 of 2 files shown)


tessig:
Does it make sense to have the whole handler instead of only the name?

Ompluscator:
> Does it make sense to have the whole handler instead of only the name?

Hi :)

I thought about it, but I could not find actual use case for that. In general, I can add that as well, but mostly I wanted to have option to know if handler is already found for request, when request is checked in Filter or Middleware.

So, would it be better to add that option as well?

bastianccm:
Hey @Ompluscator I think it's better to just provide the whole Handler struct, not just the name. Then whoever needs something can pick whatever they like.

Ompluscator:
> Hey @Ompluscator I think it's better to just provide the whole Handler struct, not just the name. Then whoever needs something can pick whatever they like.

Hi Mr Basti. Short question: would it be prefered to expose fields of Handler struct (as they are now private) and if so, which of them? If not, maybe to deliver new specific struct, which would expose some fields?

bastianccm:
Hi Mr @Ompluscator, sorry for being late, the Handler struct should have getters for all fields, right?
https://github.com/i-love-flamingo/flamingo/blob/master/framework/web/registry.go#L507
I would go with that for the time being

Ompluscator:
> Hi Mr @Ompluscator, sorry for being late, the Handler struct should have getters for all fields, right?
> https://github.com/i-love-flamingo/flamingo/blob/master/framework/web/registry.go#L507
> I would go with that for the time being

Ah true - thx! Then, I'll attach Handler object to response :)

bastianccm (approved):

PR #173, opened 2020-11-02
servemodule: allow port configuration with flamingo config

SchiffFlieger:
This change allows to configure the listened port with default flamingo config. I kept the `-a` parameter for the `serve` command for backward compatibility. We can think about removing it in the future, because it's possible to use the `--flamingo-config "core.serve.port: 3322"` parameter instead.

### DIFF
--- a/app.go
+++ b/app.go
@@ -354,7 +354,7 @@ func (a *servemodule) Configure(injector *dingo.Injector) {
 
 // CueConfig for the module
 func (a *servemodule) CueConfig() string {
-	return `core: serve: port: number | *3322`
+	return `core: serve: port: >= 0 & <= 65535 | *3322`
 }
 
 func serveProvider(a *servemodule, logger flamingo.Logger) *cobra.Command {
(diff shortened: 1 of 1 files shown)


app.go
```diff
@@ -349,6 +352,11 @@ func (a *servemodule) Configure(injector *dingo.Injector) {
 	})
 }
 
+// CueConfig for the module
+func (a *servemodule) CueConfig() string {
+	return `core: serve: port: number | *3322`
```
tessig, line 357:
```suggestion
	return `core: serve: port: int | *3322`
```

tessig: also set a constraint for the port range

SchiffFlieger: I added the constraint for valid port range, but it's not possible to declare it as `int`:
```
cue: marshal error at path core.serve.port: conflicting values ((int & >=0) & <=65535) and 1234 (mismatched types int and float)
```

bastianccm: I'll look into it

bastianccm (approved):

PR #174, opened 2020-11-06
framework/flamingo: add possibility to specify redis store's database

Ompluscator:
Added possibility to pass redis database for storing session data

### DIFF
--- a/framework/flamingo/sessions.go
+++ b/framework/flamingo/sessions.go
@@ -27,6 +27,7 @@ type SessionModule struct {
 	redisPassword        string
 	redisIdleConnections int
 	redisMaxAge          int
+	redisDatabase        string
 	healthcheckSession   bool
 }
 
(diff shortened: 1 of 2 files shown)


tessig:
Hi, 
that's a nice and long missing feature 🙂 

Additonally, there is an URL "magic" parser implemented in `getRedisConnectionInformation`, which already gets Host and Password from the `redis.url` config. It would be nice to also support getting the database number from there. At least for URLs like `redis://[:PASSWORD@]HOST[:PORT][/DATABASE]`.

Reference can be found here: https://metacpan.org/pod/URI::redis#DESCRIPTION

framework/flamingo/sessions.go
```diff
@@ -57,9 +58,7 @@ func (m *SessionModule) Inject(config *struct {
 	m.storeLength = int(config.StoreLength)
 	m.maxAge = int(config.MaxAge)
 	m.path = config.Path
-	m.redisHost, m.redisPassword = getRedisConnectionInformation(config.RedisURL, config.RedisHost, config.RedisPassword)
-	m.redisIdleConnections = int(config.RedisIdleConnections)
```
tessig, line 59: Did we loose this by mistake?

Ompluscator: Yes, you are right, I removed it by mistake. Now should be fine :)

tessig (approved):

PR #175, opened 2020-11-11
Added os.Exit after graceful shutdown is done

masmrlar:
A running flamingo app cannot be stopped by CRTL-C, the grafeful shutdown is only triggered, but the app does not exit. 
This is a bit tedious as the PID needs to be fetched to kill -9 the app. 

I added `os.Exit(0)` again as soon as the shutdown is done...


carstendietrich:
@masmrlar just had a look, at your fix. Indeed if you are using the prefixrouter module this fixes the hanging. The root issue here is that the module listens to the wrong events. Provided solution in PR #177. Closing here.

PR #178, opened 2020-12-02
ci: Switch to GitHub Actions

carstendietrich:
Since Travis CI becomes more and more unreliable and we have good experience in the flamingo-commerce repo with GitHub Actions we should think about a switch of CI tooling.

First draft version, which also adds checks for spelling and ineffective assignments.

The only drawback to the Travis variant is that the Go version matrix is hard coded to 1.14 / 1.* and there is no alias like oldstable / stable like in Travis.


.github/workflows/main.yml
```diff
@@ -0,0 +1,57 @@
...
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  tests:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        go: [ '1.14', '1.*' ]
```
SchiffFlieger, line 16: I think we agreed on using the version listed in go.mod (1.13 at the moment).

.github/workflows/main.yml
```diff
@@ -0,0 +1,57 @@
...
+        run: go vet ./...
+      - name: Go Fmt
+        run: |
+          fmt=$(gofmt -l .)
+          test -z $fmt || (echo "please run gofmt" ; echo $fmt ; exit 1)
+      - name: Go Lint
+        run: go run golang.org/x/lint/golint -set_exit_status $(go list ./... | grep -v test/integrationtest/projecttest/graphql)
+      - name: Spelling
+        run: go run github.com/client9/misspell/cmd/misspell -error .
+      - name: In effective assignments
```
SchiffFlieger, line 52: *Ineffective

.github/workflows/main.yml
```diff
@@ -0,0 +1,57 @@
...
+        go: [ '1.14', '1.*' ]
+    name: Tests
+    steps:
+      - uses: actions/checkout@v2
+      - name: Setup Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: ${{ matrix.go }}
+      - name: Get dependencies
+        run: go get -v -t -d ./...
```
SchiffFlieger, line 25: We could cache dependencies with [actions/cache](https://github.com/actions/cache). However, it only saves us a few seconds so I'm not sure if it's worth it. What do you think?

SchiffFlieger (changes requested):

.github/workflows/main.yml
carstendietrich: Guess we adhere to the Go core supported versions:

> Each major Go release is supported until there are two newer major releases. For example, Go 1.5 was supported until the Go 1.7 release, and Go 1.6 was supported until the Go 1.8 release.

So I will update the go.mod instead of changing things here.

.github/workflows/main.yml
carstendietrich: Yeah, let's discuss this in our next contributor's meetup.

.github/workflows/main.yml
carstendietrich: Updated go.mod to use 1.14

.github/workflows/main.yml
tessig: the download is quite fast, so let's skip this for now

SchiffFlieger (approved):

PR #184, opened 2021-02-18
ci: add nancy

jeinfeldt:
Hello,

we would benefit from using [Nancy](https://github.com/sonatype-nexus-community/nancy) to check our dependencies.
@tessig opened a PR for this previously https://github.com/i-love-flamingo/flamingo/pull/69
before `go.mod` was fully supported.

Since flamingo does not use travis anymore for CI, lets try with github actions.
Workflow suggested by: https://github.com/marketplace/actions/nancy-for-github-actions

For now I would suggest to only do this on PR, as we would otherwise break the CI on master.
After all vulnerabilities are fixed, we can adjust the behaviour.
Of course we can discuss how to integrate this workflow / naming things.

The current vulnerability is related to
https://github.com/spf13/viper/pull/957

If we decide to merge this closes https://github.com/i-love-flamingo/flamingo/pull/69


carstendietrich:
We could move this to the `Static checks` part, and we should have a look at https://docs.github.com/en/actions/reference/events-that-trigger-workflows#scheduled-events to run the dependency check on a daily basis..

carstendietrich:
@jeinfeldt please check the scheduled trigger and try to do a go get on etcd to override the vuln dependency

.github/workflows/main.yml
```diff
@@ -54,4 +54,12 @@ jobs:
       - name: Go Generate
         run: |
           go generate ./...
-          git diff --quiet || (echo 'generated go files are not up to date, check go generate, go.sum and go.mod' ; git diff ; exit 1)
\ No newline at end of file
+          git diff --quiet || (echo 'generated go files are not up to date, check go generate, go.sum and go.mod' ; git diff ; exit 1)
+      - name: Set up Go 1.x in order to write go.list file
+        uses: actions/setup-go@v2
+        with:
```
tessig, line 60: we don't need this, it's already above

.github/workflows/daily.yml
```diff
@@ -0,0 +1,21 @@
...
+  vulnerability-scan:
+    name: Vulnerability scan
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@v2
+      - name: Set up Go 1.x in order to write go.list file
+        uses: actions/setup-go@v2
+        with:
+          go-version: ^1.13
```
tessig, line 17: just take the most actual version via 1.*

tessig (changes requested):

tessig (approved):
very nice, thanks!

PR #187, opened 2021-03-05
framework/flamingo: add possibility to adapt SameSite property for cookies

Ompluscator:
Added possibility to adapt SameSite flag for cookies (so it can be supported in iFrame together with Secure=true flag).


framework/flamingo/sessions.go
```diff
@@ -124,9 +127,18 @@ func (m *SessionModule) setSessionstoreOptions(options *sessions.Options) {
 	options.Domain = ""
 	options.Path = m.path
 	options.MaxAge = m.maxAge
-	options.Secure = m.secure
+	options.Secure = true
```
tessig, line 130: why do we need this to be always true?

framework/flamingo/sessions.go
```diff
@@ -141,6 +153,7 @@ flamingo: session: {
 	cookie: {
 		secure: bool | *true
 		path: string | *"/"
+		sameSite: string | *"lax"
```
tessig, line 156:
```suggestion
		sameSite: *"lax" | "strict" | "none" | "default"
```

tessig (changes requested):

framework/flamingo/sessions.go
Ompluscator: @tessig  yup, you are right, it was pushed by mistake.

tessig (approved):
Sorry for the late reply. Looks fine and we will merge as soon as we solved our global nancy problem

PR #189, opened 2021-04-12
core/auth: Expose RawIDToken as part of OpenIDIdentity interface

carstendietrich:
In some cases you want to work with the raw JWT id token, it therefore makes sense to expose it directly via the interface. AccessToken and RefreshToken can already be reached with the TokenSource() function so this would be a nice addition.

### DIFF
--- a/core/auth/mock/identity.go
+++ b/core/auth/mock/identity.go
@@ -24,6 +24,11 @@ type (
 	}
 )
 
+// RawIDToken returns a mocked raw id token
+func (i *OIDCIdentity) RawIDToken() string {
+	return "raw-id-token"
+}
+
 // Subject getter
 func (i *Identity) Subject() string {
 	return i.Sub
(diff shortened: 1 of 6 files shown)


carstendietrich:
obsolete can be solved in the following way:
```go
func idToRawID(identity auth.Identity) (string, error) {
    rawIDgetter, ok := identity.(interface {
        RawIDToken() string
    })
    if !ok {
        token, err := identity.(oauth.OpenIDIdentity).TokenSource().Token()
        if err != nil {
            return "", fmt.Errorf("error identity does not provide a RawIDToken getter, and we failed falling back to regular tokens: %w", err)
        }
        if token, ok := token.Extra("id_token").(string); !ok {
            return "", fmt.Errorf("no id_token in token extras while falling back because there is no RawIDToken getter")
        } else {
            return token, nil
        }
    }
    rawIDtoken := rawIDgetter.RawIDToken()
    if rawIDtoken == "" {
        return "", fmt.Errorf("no id_token in request identity")
    }
    return rawIDtoken, nil
}
```

PR #200, opened 2021-07-01
all: refactor deprecated inject tags

tessig:


core/oauth/module.go
```diff
@@ -133,7 +152,7 @@ type routes struct {
 	logout   interfaces.LogoutControllerInterface
 	callback interfaces.CallbackControllerInterface
 	user     *interfaces.UserController
-	UseFake  bool `inject:"config:core.oauth.useFake"`
+	useFake  bool
```
carstendietrich, line 155: don't see that config in the Inject func of routes struct..

core/oauth/module.go
```diff
@@ -148,20 +167,20 @@ func (r *routes) Inject(
 	r.logout = logout
 	r.callback = callback
 	r.user = user
-	r.UseFake = *fake
+	r.useFake = *fake
```
carstendietrich, line 170: always false now?

carstendietrich (approved):
LGTM!

PR #202, opened 2021-07-21
core: update cue

### LINKED ISSUE #201: Update Cue library
cuelang has developed new features. Let's make them available in Flamingo, too.

current in Flamingo: cuelang.org/go@0.0.15
actual version: cuelang.org/go@0.4.0

tessig:
closes #201


tessig:
This is a breaking change and can not be done in v3

PR #206, opened 2021-08-19
Flamalyzer git workflow

domeej:
Will run the Flamalyzer in the pipeline

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - master
+      - flamalyzer-git-workflow
   pull_request:
     branches:
       - master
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -54,4 +54,9 @@ jobs:
       - name: Go Generate
         run: |
           go generate ./...
-          git diff --quiet || (echo 'generated go files are not up to date, check go generate, go.sum and go.mod' ; git diff ; exit 1)
\ No newline at end of file
+          git diff --quiet || (echo 'generated go files are not up to date, check go generate, go.sum and go.mod' ; git diff ; exit 1)
+      - name: get Flamalyzer
```
tessig, line 58: I don't think we need to get it first (see above on misspell and ineffassign)

tessig (changes requested):
We must definitely merge the cleanup branch first :)

.github/workflows/main.yml
domeej: I thought the same but I get an error if I remove it 
```
no required module provides package flamingo.me/flamalyzer; to add it: go get flamingo.me/flamalyzer
```

domeej: another idea?

tessig:
Closing because stale

PR #215, opened 2022-01-13
framework/flamingo: add config for port in redis session backend

SchiffFlieger:
With our latest setup, we have two separate environment variables for redis host and port. Unfortunately, it is not possible to join them together in the config.yml directly. With this PR, we introduce a separate config to specify the port for redis: `flamingo.session.redis.port`. 
It's still possible to include the port directly in the `flamingo.session.redis.host` config as before, so this change is not breaking.

### DIFF
--- a/framework/flamingo/sessions.go
+++ b/framework/flamingo/sessions.go
@@ -1,6 +1,7 @@
 package flamingo
 
 import (
+	"fmt"
 	"net/http"
 	"net/url"
 	"os"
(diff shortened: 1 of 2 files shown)


SchiffFlieger:
It is indeed possible to join the environment variables in the config.yaml, but you have to specify default values:
```
flamingo.session.redis.host: "%%ENV:REDIS_HOSTNAME%%redis%%:%%ENV:REDIS_PORT%%6379%%"
```

Otherwise the [regex replace in the config loader](https://github.com/i-love-flamingo/flamingo/blob/master/framework/config/loader.go#L197-L206) will not detect the parts correctly.

PR #216, opened 2022-01-16
fix(auth): add missing auth.Identity interface

bastianccm:


core/auth/oauth/oauth.go
```diff
@@ -13,6 +13,7 @@ import (
 type (
 	// Identity defines a TokenSource which is can be used to get an AccessToken vor OAuth2 flows
 	Identity interface {
+		auth.Identity
```
tessig, line 16: Please mention the breaking change in the changelog

tessig (changes requested):

PR #221, opened 2022-02-14
fix(oauth): correctly map access-token claims

bastianccm:
Currently mapping access token claims is behaving wrong. Instead of actual access token claims the mapper maps token claims.
This change unpacks the access token and maps them, but still uses the existing behaviour to not ensure breaking something


carstendietrich (approved):
lgtm

PR #273, opened 2022-07-04
feat(framework/config): module dump command

IvanMaidurov:


tessig:
Great work so far.
I think that command should not live in the web module and also detect modules of child areas.

PR #303, opened 2022-12-03
docs: Adjust several documentations. Add google example for auth module

danielpoe:


core/auth/example/multiple/go.mod
```diff
@@ -0,0 +1,61 @@
+module multiple
+
+go 1.19
+
+require (
+	flamingo.me/dingo v0.2.10
+	flamingo.me/flamingo/v3 v3.4.0
```
bastianccm, line 7: We would require ourself but in a different version, I don't think that's necessary

bastianccm (changes requested):

Issue #304, opened 2022-12-08
Opencensus Module sets X-Correlation-Id even if already present


nico-i:
### Description
Currently the open census module adds a correlation id header and value to every HTTP request. This leads to problems in our project as we already use the X-Correlation-Id header in other contexts throughout our services.

### Proposed solution
To deal with this issue I suggest to modify the corresponding [module.go](https://github.com/i-love-flamingo/flamingo/blob/master/framework/opencensus/module.go#L50) to only set the X-Correlation-Id header if it isn't already present and to instead use the existing X-Correlation-Id value if it is.

### Acknowledgements
I know that the open census module has since been deprecated but as we are currently dealing with this problem and the ETA for open census' replacement is yet to be determined, I think taking on this issue is still viable.

tessig:
closed by #310

PR #305, opened 2022-12-08
fix(opencensus): duplicate x-correlation-id

nico-i:
This fixes issue #304.


tessig:
@nico-i There seems to be some conflict with the master branch. Would you be so nice and rebase your branch on our master?

PR #306, opened 2022-12-12
fix(cache): do not rewrite trace id

IvanMaidurov:


core/cache/httpFrontend.go
```diff
@@ -104,7 +124,7 @@ func (hf *HTTPFrontend) load(ctx context.Context, key string, loader HTTPLoader,
 	defer span.End()
 
 	data, err := hf.Do(key, func() (res interface{}, resultErr error) {
-		ctx, fetchRoutineSpan := trace.StartSpan(context.Background(), "flamingo/cache/httpFrontend/fetchRoutine")
+		ctx, fetchRoutineSpan := trace.StartSpan(ctx, "flamingo/cache/httpFrontend/fetchRoutine")
```
IvanMaidurov, line 127: perhaps we still should start new context here, but span should derivate from previous

bastianccm (approved):

PR #308, opened 2022-12-16
feat(zap): silent zap logging

### LINKED ISSUE #311: Add ApplicationOption to specify logger (and others?)
Like `WithCoreLogger`, which replaces zap.Module in https://github.com/i-love-flamingo/flamingo/blob/5909a729b3075163e0e961f702de47a482a6dfbb/app.go#L131

IvanMaidurov:
Logger will write to output only if error occurred during request

Closes #311

### DIFF
--- a/app.go
+++ b/app.go
@@ -13,16 +13,17 @@ import (
 	"time"
 
 	"flamingo.me/dingo"
+	"github.com/spf13/cobra"
+	"go.opencensus.io/plugin/ochttp"
+
 	"flamingo.me/flamingo/v3/core/runtime"
-	"flamingo.me/flamingo/v3/core/zap"
+	"flamingo.me/flamingo/v3/core/silentlogger"
 	"flamingo.me/flamingo/v3/framework"
 	"flamingo.me/flamingo/v3/framework/cmd"
 	"flamingo.me/flamingo/v3/framework/config"
 	"flamingo.me/flamingo/v3/framework/flamingo"
 	"flamingo.me/flamingo/v3/framework/opencensus"
 	"flamingo.me/flamingo/v3/framework/web"
-	"github.com/spf13/cobra"
-	"go.opencensus.io/plugin/ochttp"
 )
 
 type (
@@ -128,7 +129,7 @@ func NewApplication(modules []dingo.Module, options ...ApplicationOption) (*Appl
 
 	modules = append([]dingo.Module{
 		new(framework.InitModule),
-		new(zap.Module),
+		new(silentlogger.Module),
 		new(runtime.Module),
 		new(cmd.Module),
 	}, modules...)
--- a/app.go
+++ b/app.go
@@ -17,7 +17,7 @@ import (
 	"go.opencensus.io/plugin/ochttp"
 
 	"flamingo.me/flamingo/v3/core/runtime"
-	"flamingo.me/flamingo/v3/core/silentlogger"
+	"flamingo.me/flamingo/v3/core/zap"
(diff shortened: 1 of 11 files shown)


IvanMaidurov:
easy test with no parallel requests but with parallel writings:

```
name              old time/op    new time/op    delta
Log/spam_test-10     2.28s ± 1%     0.45s ±77%   -80.29%  (p=0.000 n=10+10)

name              old alloc/op   new alloc/op   delta
Log/spam_test-10    95.0MB ± 1%   216.7MB ± 0%  +128.20%  (p=0.000 n=9+10)

name              old allocs/op  new allocs/op  delta
Log/spam_test-10     1.33M ± 1%     1.39M ± 1%    +4.43%  (p=0.000 n=10+10)
```

100 parallel requests, each. with 300 parallel writes:

```
name              old time/op    new time/op    delta
Log/spam_test-10     2.02s ± 1%     2.13s ± 1%     +5.36%  (p=0.000 n=10+10)

name              old alloc/op   new alloc/op   delta
Log/spam_test-10     198MB ± 1%   11409MB ± 0%  +5657.54%  (p=0.000 n=10+9)

name              old allocs/op  new allocs/op  delta
Log/spam_test-10     3.20M ± 0%     6.07M ± 0%    +89.84%  (p=0.000 n=10+10)
```

app.go
```diff
@@ -128,7 +129,7 @@ func NewApplication(modules []dingo.Module, options ...ApplicationOption) (*Appl
 
 	modules = append([]dingo.Module{
 		new(framework.InitModule),
-		new(zap.Module),
```
bastianccm, line 131: I would not change the default here, but instead do that with #311

core/silentlogger/module.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+)
+
+// Configure the ZAP logger as flamingo.Logger (in JSON mode kibana compatible)
+func (m *Module) Configure(injector *dingo.Injector) {
+	registry := new(LoggingContextRegistry)
+
+	injector.Bind(registry).AsEagerSingleton()
+
+	injector.Bind(new(flamingo.Logger)).ToProvider(getSilentLogger)
```
bastianccm, line 26: `getSilentLogger` is quite a "heavy" implementation, can this be a performance issue if we call the factory for every logger instance?

core/silentlogger/module.go
```diff
@@ -0,0 +1,71 @@
...
+	Module struct{}
+
+	shutdownEventSubscriber struct {
+		logger flamingo.Logger
+	}
+)
+
+// Configure the ZAP logger as flamingo.Logger (in JSON mode kibana compatible)
+func (m *Module) Configure(injector *dingo.Injector) {
+	registry := new(LoggingContextRegistry)
```
bastianccm, line 22: This is not necessary, rather reference them in the bindings `Bind(new(LoggingContextRegistry))`...

bastianccm (changes requested):

core/silentlogger/module.go
IvanMaidurov: in line 27 I used this reference in ```flamingo.BindEventSubscriber(injector).To(registry)```, will EagerSingleton still be referenced if I create two instances?

core/silentlogger/module.go
IvanMaidurov: I guess no really big differences, all info is getting cleared every request end. So it either lives in one instance, or spread across different instances. By itself it's not heavy in my opinion

bastianccm (approved):

PR #338, opened 2023-05-15
feat(core/auth/oidc): Add support of `AuthCodeOption` in the token exchange during callback

carstendietrich:
Currently, it is only possible to modify the authorization URL using the `AuthCodeOptioner` interface. Since the Exchange function also supports this pattern we should also pass the registered `AuthCodeOptioner` to it.

For example, both my authorization and my token retrieval endpoint need some custom URL param to work. Only the one for the auth endpoint would currently be possible.

### DIFF
--- a/core/auth/oauth/oidc.go
+++ b/core/auth/oauth/oidc.go
@@ -467,7 +467,15 @@ func (i *openIDIdentifier) Callback(ctx context.Context, request *web.Request, r
 		return i.responder.ServerError(err)
 	}
 
-	oauth2Token, err := oauthConfig.Exchange(ctx, code)
+	options := make([]oauth2.AuthCodeOption, 0, len(i.authcodeOptions))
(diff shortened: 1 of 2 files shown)


core/auth/oauth/oidc.go
```diff
@@ -467,7 +467,15 @@ func (i *openIDIdentifier) Callback(ctx context.Context, request *web.Request, r
 		return i.responder.ServerError(err)
 	}
 
-	oauth2Token, err := oauthConfig.Exchange(ctx, code)
+	options := make([]oauth2.AuthCodeOption, 0, len(i.authcodeOptions))
+	for _, o := range i.authcodeOptions {
```
carstendietrich, line 471: we should drop this for loop since this is only used for setting the scope param which is not needed in the token endpoint call..

KarolNowakk (approved):
Beautiful

PR #353, opened 2023-08-18
ci: force on latest go version & cleanup jobs

tessig:
remove matrix strategy from jobs where we only check one go version


carstendietrich (approved):
lgtm

PR #403, opened 2024-05-02
feat(auth/oidc): Use BadRequest response for state mismatches, enhance logging with context

carstendietrich:
Using HTTP status code `500` for all issues occurring during the OIDC flow might not be the best way. Therefore changing at least the errors that can occur during the callback e.g. tinkered/missing state param as HTTP status `400`.

Additionally extending the web responder with error funcs that also take the ctx to enhance the logging for better traceability.

### DIFF
--- a/core/auth/oauth/oidc.go
+++ b/core/auth/oauth/oidc.go
@@ -443,12 +443,12 @@ func (i *openIDIdentifier) Authenticate(ctx context.Context, request *web.Reques
 
 	authConfig, err := i.config(request)
 	if err != nil {
-		return i.responder.ServerError(err)
+		return i.responder.ServerErrorWithContext(ctx, err)
 	}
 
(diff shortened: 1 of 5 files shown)


tessig (approved):
lgtm

core/auth/oauth/oidc.go
```diff
@@ -109,6 +109,11 @@ var (
 
 		return ok
 	}
+
+	errNoStateInRequest = errors.New("no state in request")
+	errStateMismatch    = errors.New("state mismatch")
+	errNoIDTokenClaim   = errors.New("claim id_token missing")
+	errGeneric          = errors.New("OpenID Connect error")
```
IvanMaidurov, line 116: Should we make those exported to be able to handle them?

carstendietrich: they are only passed to the responder to render the error pages, so there is currently no place where you could check for these errors outside of this package.

PR #412, opened 2024-07-05
feat(log): Introduce Trace log level

svedatbegar:
The logger implementation/interface within Flamingo does not support Trace level logging. This PR adds that functionality.

### DIFF
--- a/core/silentzap/levels_standard.go
+++ b/core/silentzap/levels_standard.go
@@ -3,6 +3,4 @@
 package silentzap
 
 // allowedLevels returns allowed levels for the standard logger.
-func allowedLevels() string {
-	return `*"Debug" | "Info" | "Warn" | "Error" | "DPanic" | "Panic" | "Fatal"`
-}
+const allowedLevels = `*"Debug" | "Info" | "Warn" | "Error" | "DPanic" | "Panic" | "Fatal"`
--- a/core/silentzap/levels_standard.go
+++ b/core/silentzap/levels_standard.go
@@ -1,0 +1,8 @@
+//go:build !tracelog
+
+package silentzap
+
+// allowedLevels returns allowed levels for the standard logger.
+func allowedLevels() string {
+	return `*"Debug" | "Info" | "Warn" | "Error" | "DPanic" | "Panic" | "Fatal"`
+}
--- a/core/silentzap/levelencoder.go
+++ b/core/silentzap/levelencoder.go
@@ -1,0 +1,25 @@
+package silentzap
+
+import (
+	"fmt"
+
+	"go.uber.org/zap/zapcore"
+)
+
+func capitalLevelEncoder(l zapcore.Level, enc zapcore.PrimitiveArrayEncoder) {
+	if l == logLevels["Trace"] {
+		enc.AppendString("TRACE")
+		return
+	}
+
+	zapcore.CapitalLevelEncoder(l, enc)
+}
+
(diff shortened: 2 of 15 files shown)


tessig:
Hi @svedatbegar 
 
Thanks for implementing this feature. We appreciate your effort and the proposed changes.

Please also adjust the `CueConfig` functions of `zap.Module` and silentzap.Module` to accept `Trace` as log level.

However, the change to the interface `flamingo.Logger` is a breaking change and, strictly seen, this would require us to increase the major version to v4, which is not something we are planning to do in the near future.

I guess it is very unlikely that projects have their own logger implementations in place, but we must still consider it since we strive to maintain backward compatibility as much as possible.

I'd like to pull this topic into a bigger round. We have a maintainers meeting this Thursday and will discuss it there. I will be happy to forward you the invitation if you are interested in taking part, too.

svedatbegar:
Hi @tessig

Thanks for taking a look at the PR. I've made the changes you requested.

If it's necessary to bump the major version for the logger, I'm assuming it could take a while to introduce this feature as you mentioned. But I must add that since trace level logging does not exist in flamingo right now, we'll have to implement or bring in our own logger that has this functionality. We'll have to discuss with the team on how to proceed.

I'd be glad to join the round to discuss this issue.

svedatbegar:
Hello again @tessig ,

I've made the updates you requested. I kept the trace implementations in place while locking cue config updates and the logger interface behind build tags as you suggested. As far as I understand, there are no problems with adding more functionality to the implementations as long as we don't touch the interface.

1 more thing to keep an eye on would be the testing of the trace level logs, as the tests must be copied for them or maybe should be handled differently, but I'm open to suggestions.

tessig:
Very nice work @svedatbegar 🙇 

I agree, the example tests are a problem, because godoc does not support build tags so far: https://github.com/golang/go/issues/3398

With zap.Logger's Trace function being tested, I would be fine with having the Trace not in the ExampleTest for now, or a duplicated ExampleTest (which will not render in godoc for now).

Also, I feel the need that we should mention the tracelog feature and build tag in the documentation. Perhaps we can extend the logging section in `docs/1. Flamingo Basics/5. Flamingo Production.md`. But that can also be done in a separate PR.

core/silentzap/levels_standard.go
```diff
@@ -0,0 +1,8 @@
+//go:build !tracelog
+
+package silentzap
+
+// allowedLevels returns allowed levels for the standard logger.
+func allowedLevels() string {
+	return `*"Debug" | "Info" | "Warn" | "Error" | "DPanic" | "Panic" | "Fatal"`
```
carstendietrich, line 7: Is there a reason we went for a function? Would prefer a const instead..

tessig:
We do not duplicate the example tests. It's not worth the effort.
Let's just add a commented line, eg
```
// logger.Trace("my", "log", "args") // this is possible with build tag `tracelog`
```

carstendietrich (changes requested):
overall the changes look good to me, would only move the allowed levels to a const instead of a function. After this adjustment, we could merge the changes.

core/silentzap/levels_standard.go
svedatbegar: Perfectly reasonable request. Making the change.

carstendietrich (approved):

tessig (approved):

tessig:
Thanks for your contribution, it will be included in v3.9.0. We tidy up documentation and then tag.

PR #415, opened 2024-07-18
chore: disable mockery version string

tessig:


.mockery.yaml
```diff
@@ -2,3 +2,4 @@ inpackage: False
 testonly: False
```
carstendietrich, line 2: lets lower case all bool values..

carstendietrich (approved):

PR #424, opened 2024-08-12
feat(flamingo): add optional attribute to configure a session key prefix for REDIS (flamingo.session.redis.keyPrefix)

mkrill:
…


framework/flamingo/sessions.go
```diff
@@ -55,6 +57,7 @@ func (m *SessionModule) Inject(config *struct {
 	Path                 string  `inject:"config:flamingo.session.cookie.path"`
 	RedisURL             string  `inject:"config:flamingo.session.redis.url"`
 	RedisHost            string  `inject:"config:flamingo.session.redis.host"`
+	RedisKeyPrefix       string  `inject:"config:flamingo.session.redis.keyPrefix,optional"`
```
karstenkoehler, line 60: Could you add the key to the `CueConfig` method for validation and also document it in the `SessionReadme.md` file?

karstenkoehler (changes requested):
Hi @mkrill, thanks for your contribution. Could you please have a look at the changes, which I proposed?

framework/flamingo/sessions.go
mkrill: Hello Karsten, 
thx for your quick feedback. I updated ReadmeSession.md and added an entry in CueConfig(). Is there any reason, why the `?` for indicating optional parameters wasn't used so far?

mkrill: @karstenkoehler Regarding `CueConfig()` there is a second option: `keyPrefix: string | ""`. Please let me know, if you prefer the other one.

framework/flamingo/sessions.go
```diff
@@ -76,6 +79,10 @@ func (m *SessionModule) Inject(config *struct {
 	m.redisClusterMode = config.RedisClusterMode
 	m.healthcheckSession = config.CheckSession
 
+	if strings.TrimSpace(config.RedisKeyPrefix) != "" {
+		m.redisKeyPrefix = &config.RedisKeyPrefix
```
karstenkoehler, line 83: @mkrill We should also trim the value here before assigning it.

mkrill: Done. Might also make sense for other parameters in the future, probably.

karstenkoehler (approved):

PR #442, opened 2024-09-18
feat: decouple opencensus from core

tessig:
This is in preparation of the opentelemetry drop-in replacement


carstendietrich (approved):
LGTM

PR #485, opened 2025-03-03
Add tests for event dispatch when running commands and graceful shutdown

kgrigorev:
This PR adds some tests to cover command execution and triggering startup and shutdown events. In order to enable testing it introduces some minor refactoring of error handling and graceful shutdown

### DIFF
--- a/app_test.go
+++ b/app_test.go
@@ -1,0 +1,250 @@
+package flamingo_test
+
+import (
+	"context"
+	"fmt"
+	"net"
+	"os"
+	"sync"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"flamingo.me/dingo"
+	"github.com/spf13/cobra"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"flamingo.me/flamingo/v3"
+	"flamingo.me/flamingo/v3/framework/cmd"
+	framework "flamingo.me/flamingo/v3/framework/flamingo"
+)
+
+type NotifyFunc func(ctx context.Context, event framework.Event)
+
+func (nf NotifyFunc) Notify(ctx context.Context, event framework.Event) {
+	nf(ctx, event)
+}
+
+func TestCmdEventsTriggeredProperly(t *testing.T) { //nolint:paralleltest // due to dingo.Singleton
+	assertStartupAndShutdownOnce := func(startupEventCount, shutdownEventCount int32) {
+		assert.Equal(t, int32(1), startupEventCount, "startupEventCount should be 1")
+		assert.Equal(t, int32(1), shutdownEventCount, "shutdownEventCount should be 1")
+	}
+
+	tests := []struct {
+		name string
+		args string
+		want func(startupEventCount, shutdownEventCount int32)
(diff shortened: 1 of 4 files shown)


app_test.go
```diff
@@ -0,0 +1,250 @@
...
+)
+
+type NotifyFunc func(ctx context.Context, event framework.Event)
+
+func (nf NotifyFunc) Notify(ctx context.Context, event framework.Event) {
+	nf(ctx, event)
+}
+
+func TestCmdEventsTriggeredProperly(t *testing.T) { //nolint:paralleltest // due to dingo.Singleton
+	assertStartupAndShutdownOnce := func(startupEventCount, shutdownEventCount int32) {
```
tessig, line 30: this function should receive the `testing.T` from the subtest (as first arg and be `t.Helper()`)

app_test.go
```diff
@@ -0,0 +1,250 @@
...
+		assert.Equal(t, int32(1), startupEventCount, "startupEventCount should be 1")
+		assert.Equal(t, int32(1), shutdownEventCount, "shutdownEventCount should be 1")
+	}
+
+	tests := []struct {
+		name string
+		args string
+		want func(startupEventCount, shutdownEventCount int32)
```
tessig, line 38: seems the `want` is not used at all

app_test.go
```diff
@@ -0,0 +1,250 @@
...
+		process, err := os.FindProcess(pid)
+		require.NoError(t, err)
+
+		err = process.Signal(os.Interrupt)
+		require.NoError(t, err)
+	}
+}
+
+func TestGracefulShutdown(t *testing.T) { //nolint:paralleltest // due to dingo.Singleton
+	assertShutdownOnce := func(shutdownEventCount int32) {
```
tessig, line 122: this also should receive `t` and be `t.Helper()`

app_test.go
```diff
@@ -0,0 +1,250 @@
...
+
+	tests := []struct {
+		name                string
+		args                string
+		onServerStartup     func()
+		onShutdown          func()
+		insideCommandRun    func(cmd *cobra.Command, args []string)
+		insideCommandRunE   func(cmd *cobra.Command, args []string) error
+		wantErr             assert.ErrorAssertionFunc
+		assertShutdownCount func(shutdownEventCount int32)
```
tessig, line 134: is this used?

app_test.go
```diff
@@ -0,0 +1,250 @@
...
+	"github.com/spf13/cobra"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"flamingo.me/flamingo/v3"
+	"flamingo.me/flamingo/v3/framework/cmd"
+	framework "flamingo.me/flamingo/v3/framework/flamingo"
+)
+
+type NotifyFunc func(ctx context.Context, event framework.Event)
```
tessig, line 23: I would make it private

tessig (changes requested):

app_test.go
kgrigorev: get lost during some edits, restored it

tessig (approved):

PR #503, opened 2025-05-14
feat(healthcheck): add measured status healthcheck interface

IvanMaidurov:


core/healthcheck/interfaces/controllers/healthcheck.go
```diff
@@ -29,19 +35,46 @@ type (
 	}
 )
 
+const (
+	statusMeasureChecksName   = "flamingo/status/checks"
+	statusMeasureFailuresName = "flamingo/status/failures"
```
tessig, line 40: should we use something ending in total? see https://prometheus.io/docs/practices/naming/#metric-and-label-naming

core/healthcheck/domain/healthcheck/status.go
```diff
@@ -1,6 +1,13 @@
...
-}
+type (
+	// Status check interface
+	Status interface {
+		Status() (alive bool, details string)
+	}
+
+	MeasuredStatus interface {
+		Status
+		ServiceName() string
```
tessig, line 11: A status is not necessarily connected to a service. I would just go for `Name`

core/healthcheck/interfaces/controllers/healthcheck.go
IvanMaidurov: done

core/healthcheck/domain/healthcheck/status.go
```diff
@@ -1,6 +1,13 @@
...
-type Status interface {
-	Status() (alive bool, details string)
-}
+type (
+	// Status check interface
+	Status interface {
+		Status() (alive bool, details string)
+	}
+
+	MeasuredStatus interface {
```
kgrigorev, line 9: since it is a public API it may be worth adding a comment with the description why this interface is needed

PR #506, opened 2025-05-22
chore: migrate golangci-lint to v2

kgrigorev:


core/silentzap/logger.go
tessig, line 1: why do we remove the explicit function calls of logger?

core/zap/logger.go
tessig, line 1: why do we remove the explicit function calls of logger?

kgrigorev: They were producing:  QF1008: could remove embedded field "Logger" from selector (staticcheck)

core/silentzap/logger.go
kgrigorev: https://github.com/i-love-flamingo/flamingo/pull/506/files/1d2e5a457d031d63859113c583db90f2a0160a38#r2102758315

tessig (approved):

Issue #511, opened 2025-06-11
Zap Logger showing wrong source - configurable AddCallerSkip


jaimecasero:
All My logs using flamingo show source field with zap logger library as in 

`{"level":"DEBUG","@timestamp":"2025-06-11T11:31:35.900+0200","source":"zap/logger.go:116","message":"no valid identity found in current context/request","area":"root","traceID":"788603bf132a4bfe96f2fa638f22b39a","spanID":"5a87b82339334ec9","method":"POST","path":"/graphql","session":"a157789ba37d48ecf4662ef6b30f585a17a762641aa939ef14126102860fef8c"}
`

The AddCallerSkip option is hardcoded to "1" resulting in this behavior.

The number of level to skip should be configurable from config file like logsession or json format, allowing each application to select how many levels to skip in the stack depending on wrapping layers...

https://github.com/i-love-flamingo/flamingo/blob/master/core/zap/module.go#L158

I can quickly contribute a MR for this issue, as I have already tried this locally

tessig:
Thanks for pointing this out! I believe that is a relic of v2, which went unnoticed for years 🙈

tessig:
resolved by #512

PR #512, opened 2025-06-11
feat: zap logger caller skip configurable

jaimecasero:
as described in https://github.com/i-love-flamingo/flamingo/issues/511

Make zap logger caller skip configurable so source field shows the actual stack level from users application

### DIFF
--- a/core/zap/module.go
+++ b/core/zap/module.go
@@ -25,6 +25,7 @@ type (
 		fieldMap           map[string]string
 		logSession         bool
 		callerEncoder      zapcore.CallerEncoder
+		callerskip         int
 	}
 
 	shutdownEventSubscriber struct {
@@ -68,6 +69,7 @@ func (m *Module) Inject(config *struct {
 	FieldMap           config.Map `inject:"config:core.zap.fieldmap,optional"`
 	LogSession         bool       `inject:"config:core.zap.logsession,optional"`
 	CallerEncoder      string     `inject:"config:core.zap.encoding.caller,optional"`
+	CallerSkip         int        `inject:"config:core.zap.callerskip,optional"`
 }) {
 	m.area = config.Area
 	m.json = config.JSON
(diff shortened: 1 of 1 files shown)


jaimecasero:
image
logs are showing proper source now

jaimecasero:
arguably caller skip default value should be 2 already, since 1 will only show flamingo logger. not sure under which circumstances caller skip 1 could be useful for anyone using flamingo lib

core/zap/module.go
```diff
@@ -208,6 +217,7 @@ core: zap: {
 	encoding: {
 		caller: *"%s" | "%s" | "%s"
 	}
+	callerskip: float64 | *1
```
tessig, line 220:
```suggestion
	callerskip: float64 | *2
```

core/zap/module.go
```diff
@@ -80,6 +82,13 @@ func (m *Module) Inject(config *struct {
 	m.logSession = config.LogSession
 	m.callerEncoder = callerEncoders[ZapCallerEncoderShort]
 
+	//if not provided set 1 as default
+	m.callerskip = 1
```
tessig, line 86: yes, as you mentioned, let's set the default value to 2

```suggestion
	m.callerskip = 2
```

tessig (changes requested):
Thanks for providing this PR!
Let's set the default to 2

jaimecasero:
@tessig thanks for checking, i just fixed some lint issues, hoping to merge soon. what is your release cycle? when are contributions added to next release? let me know if im missing anything to push this forward, im hoping to fix my logs with next version without having to fork

core/zap/module.go
```diff
@@ -80,6 +82,13 @@ func (m *Module) Inject(config *struct {
 	m.logSession = config.LogSession
 	m.callerEncoder = callerEncoders[ZapCallerEncoderShort]
 
+	// if not provided set 1 as default
```
tessig, line 85:
```suggestion
	// if not provided set 2 as default
```

tessig:
@jaimecasero Thanks for the changes. I quickly corrected that one comment and will merge today.
We have no fixed release cycle. I quickly double-check with the other maintainers, but think that we can release directly today

tessig:
@jaimecasero The release is out 🚀 
https://github.com/i-love-flamingo/flamingo/releases/tag/v3.15.0
Thanks again!

PR #524, opened 2025-08-05
feat(session): add username support for redis

IvanMaidurov:


framework/flamingo/sessions.go
```diff
@@ -224,6 +233,7 @@ flamingo: session: {
 	redis: {
 		url: string | *""
 		host: string | *"redis"
+		username?: string & !=""
```
tessig, line 236: do we want to go away from the usual `username: string | *""`?
After injection, it will be an empty string anyway

framework/flamingo/sessions.go
```diff
@@ -73,7 +75,12 @@ func (m *SessionModule) Inject(config *struct {
 	m.storeLength = int(config.StoreLength)
 	m.maxAge = int(config.MaxAge)
 	m.path = config.Path
-	m.redisHost, m.redisPassword, m.redisDatabase = getRedisConnectionInformation(config.RedisURL, config.RedisHost, config.RedisPassword, config.RedisDatabase)
+
+	m.redisHost = getRedisHost(config.RedisURL, config.RedisHost)
+	m.redisUsername = getRedisUsername(config.RedisURL, config.RedisUsername)
+	m.redisPassword = getRedisPassword(config.RedisURL, config.RedisPassword)
+	m.redisDatabase = getRedisDatabase(config.RedisURL, config.RedisDatabase)
```
tessig, line 82: With this function split, we parse the URL 4 times. No idea if it is an expensive operation

framework/flamingo/sessions.go
```diff
@@ -255,24 +265,68 @@ func (m *SessionModule) FlamingoLegacyConfigAlias() map[string]string {
...
+	return redisPassword
+}
+
+func getRedisHost(redisURL string, redisHost string) string {
+	if redisURL == "" {
+		return redisHost
+	}
+
+	parsedRedisURL, err := url.Parse(redisURL)
+	if err != nil {
```
tessig, line 310: Now that we are touching this, I think this is worth not being silenced.
When I configure an URL and have a typo, this secretly falls back to the host which is not wanted and searching for the error is hard

framework/flamingo/sessions.go
IvanMaidurov: yeah, moved parsing out

PR #576, opened 2026-04-16
feat: expose helper method to obtain http.Handler from the application

kgrigorev:
The Go standard library provides convenient web server testing facilities via httptest.Server. This PR introduces a helper method to obtain an http.Handler from the configured flamingo.Application. It also makes the Application itself an http.Handler by implementing ServeHTTP method. This allows following usage in tests:

```go
	app, err := flamingo.NewApplication(modules, ...)
	if err != nil { ... }

	handler, err := app.HTTPHandler()
	if err != nil { ... }

	testServer := httptest.NewServer(handler) // or even httptest.NewServer(app)
	defer testServer.Close()

	requestURL := fmt.Sprintf("%s/hello", testServer.URL)
	response, err := testServer.Client().Get(requestURL)
	if assert.NoError(t, err) {
		...
	}
```

### DIFF
--- a/app.go
+++ b/app.go
@@ -41,6 +41,8 @@ const (
 var (
 	// ErrAppRun is returned when an error occurs during app.Run()
 	ErrAppRun = errors.New("app run error")
+
+	_ http.Handler = new(Application)
 )
 
 type (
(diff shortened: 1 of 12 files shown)


tessig:
Thanks, this is a useful enhancement. Please address the failing tests and lints before we can merge.

PR #587, opened 2026-06-30
fix(framework/web): do not expose partial rendering errors

tessig:
Any client or attacker can freely set the `X-Partial` header. We should not expose the real template engine error and only respond with a general failure.

A classical example for an attack can be: `X-Partial: (select(0)from(select(sleep(15)))v)/*'\''+(select(0)from(select(sleep(15)))v)+'\''"+(select(0)from(select(sleep(15)))v)+"'`

### DIFF
--- a/framework/web/result.go
+++ b/framework/web/result.go
@@ -357,17 +357,23 @@ func (r *RenderResponse) Apply(c context.Context, w http.ResponseWriter) error {
 			for k, v := range content {
 				buf, err := io.ReadAll(v)
 				if err != nil {
-					return err
+					w.WriteHeader(http.StatusInternalServerError)
+
+					return ErrPartialRenderingFailed
 				}
 				result[k] = string(buf)
 			}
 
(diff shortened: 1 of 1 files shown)


kgrigorev (approved):

framework/web/result.go
```diff
@@ -344,7 +348,9 @@ func (r *RenderResponse) Apply(c context.Context, w http.ResponseWriter) error {
 		if partials := req.Request().Header.Get("X-Partial"); partials != "" && ok {
 			content, err := partialRenderer.RenderPartials(c, r.Template, r.Data, strings.Split(partials, ","))
 			if err != nil {
-				return err
+				w.WriteHeader(http.StatusInternalServerError)
```
kgrigorev, line 351: Looks like we don't invoke WriteHeader in other error cases in this function; I would assume it is handled down the line

tessig: good point, the status is set at the very top when an error bubbles up, but the metric is collected earlier in `framework/web/filter/request_metrics_filter.go:84` and will be wrong if we do not write the header.
I will touch the other returns of the function as well

